PR #88, opened 2024-08-09
Enable to choose a camera device in UI

samoylenkodmitry:
image

## Summary by Sourcery

Introduce a feature in the UI to allow users to choose a camera device from a dropdown menu. Enhance the `webcam_preview` function to support camera selection by passing the selected camera index.

New Features:
- Add UI elements to allow users to select a camera device from a dropdown menu.

Enhancements:
- Modify the `webcam_preview` function to accept a camera index parameter, enabling the selection of different camera devices.

PR #115, opened 2024-08-10
Revert "Enable to choose a camera device in UI"

hacksider:
Reverts hacksider/Deep-Live-Cam#88
Will put this on experimental

## Summary by Sourcery

Revert the feature that allowed users to choose a camera device in the UI, reverting to the default camera selection.


giusparsifal:
> Reverts #88 Will put this on experimental
> 
> ## Summary by Sourcery
> Revert the feature that allowed users to choose a camera device in the UI, reverting to the default camera selection.

Hello, I don't understand if this feature is already possibile or not. It would be great and, by the way, you did an excellent work compared to other deep swap live cam!
Have you consider to add also FACE MASK TYPES as box, occlusion, region etc?
Thanks anyway, I'll thank you to paypal too!

PR #179, opened 2024-08-11
Update onnxruntime-gpu version in readme

munimdev:
The version 1.16.3 cannot be installed anymore but the readme file still lists 1.16.3 if one wishes tot use CUDA. This version (1.18.0) also matches the one in requirements.txt file

## Summary by Sourcery

Update the onnxruntime-gpu version in the README to 1.18.0 to ensure consistency with the requirements.txt file and address the unavailability of version 1.16.3.

Documentation:
- Update the onnxruntime-gpu version in the README from 1.16.3 to 1.18.0 to match the version in requirements.txt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ pip install -r requirements.txt
 
 ```
 pip uninstall onnxruntime onnxruntime-gpu
-pip install onnxruntime-gpu==1.16.3
+pip install onnxruntime-gpu==1.18.0
 
 ```
 


teddybear082:
I'm not sure this change is right. I tried running with cuda on the originally installed 1.18 version (which is in requirements.txt so installed by default) and it did not work, I just got a box and noise around my face.  Following the readme to uninstall that and install the earlier version of onnxruntime fixed my issue.  Also, there would be no need for the readme instructions at all if the version worked that was installed by default anyway.  So if this change worked for you it may not be universal.  I am on windows.

munimdev:
I see. If it's not universal, we can close this PR.

PR #329, opened 2024-08-15
Added virtual camera output and fetching of input camera devices with names using pygrabber on windows and linux

bit-wrangler:
The primary change in this PR is to allow the output of video to a virtual camera using pyvirtualcam package. This should work on Windows and Linux. IIRC, this may not work on osx - so, I put it behind a toggle. I refactored the live preview loop contents to be a function so that the loop can be conditionally place inside a "with virtualcam" scope.
I increased the height of the window to fit the extra toggle, plus space things out just a little bit.
I've also used the pygrabber package to retrieve input camera options on windows and linux with their names, which IMO is better than seeing "Camera 0", "Camera 1", etc.

## Summary by Sourcery

Introduce virtual camera output functionality and enhance camera device selection by displaying device names using pygrabber. Refactor the live preview loop for better modularity and adjust UI layout to support new features.

New Features:
- Add support for outputting video to a virtual camera using the pyvirtualcam package on Windows and Linux.
- Implement fetching of input camera devices with their names using the pygrabber package on Windows and Linux.

Enhancements:
- Refactor the live preview loop into a separate function to allow conditional execution within a virtual camera context.
- Increase the height of the UI window to accommodate additional UI elements and improve spacing.


bit-wrangler:
I merged and resolved conflicts from upstream. I've also added a README note for new functionality.

bit-wrangler:
Down the line, it would be nice to add audio capture and audio output with a delay to match audio to video. And an even further goal might be to add audio voice mapping between the audio input and output.

frostworx:
Hi,
thank you for this interesting PR!
I've read that pyvirtualcam uses v4l2loopback, which should _(in theory)_ make it possible to populate the new v4l device as usb webcam via usb uvc gadget device _(so any device attached to it would see a regular ready to use webcam)_

I just tried the experimental branch, but it crashes with the following trace on my (Arch Linux) system:

```
python run.py --execution-provider cuda
Traceback (most recent call last):
  File "/media/nvme/ai/Deep-Live-Cam-experimental/run.py", line 3, in <module>
    from modules import core
  File "/media/nvme/ai/Deep-Live-Cam-experimental/modules/core.py", line 21, in <module>
    import modules.ui as ui
  File "/media/nvme/ai/Deep-Live-Cam-experimental/modules/ui.py", line 9, in <module>
    from pygrabber.dshow_graph import FilterGraph
  File "/media/nvme/ai/deep-live-cam/lib/python3.10/site-packages/pygrabber/dshow_graph.py", line 33, in <module>
    from ctypes import POINTER, byref, cast, create_unicode_buffer, pointer, windll, wstring_at
ImportError: cannot import name 'windll' from 'ctypes' (/usr/lib/python3.10/ctypes/__init__.py)
```

Likely the cause is

`from pygrabber.dshow_graph import FilterGraph`

(or better DirectShow -> pygrabber generally) being windows only

No big deal, just wanted to let you know

PR #621, opened 2024-09-23
Significantly improve video resolution/quality using ESPCN_x4 model

saleweaver:
To use the resolution enhancer (super resolution), simply add super_resolution to the frame processors. It's a huge difference in quality.

Scale factor is from 2x-4x

    python run.py -t "<VIDEO>" -s "<SOURCE>" -o ./out.mp4 --video-quality 18 --keep-fps --headless  --frame-processor face_swapper super_resolution

In addition to sourcery's output, the -r option can be used to select the model for upscaling (2,3 or 4x)


saleweaver:
Hey @hacksider, does it work for you?

PR #678, opened 2024-10-02
update create_webcam_preview

highbuyer:
update create_webcam_preview

## Summary by Sourcery

Update the create_webcam_preview function to use a specific camera index with the DirectShow backend, allowing for more precise control over the webcam selection.

Enhancements:
- Modify the webcam preview function to use a specific camera index with DirectShow backend instead of the default camera.


hacksider:
Please make the IP dynamic as to ensure that it wont pose any security issues. As well as to keep the local IP dynamic as well :)

PR #878, opened 2025-01-06
Ubuntu2204 webcam mode bugfix

renmengqisheng:
In my environment (OS: Ubuntu22.04, Python: 3.10.16, CUDA: 11.8, GPU: RTX 4070 laptop), the value of `PREVIEW.winfo_width()` is 1, which causes the `fit_image_to_size` function to have a value of 0 when calculating `new_size` and `cv2.resize(image, dsize=new_size)` crashes, as shown in the following figure.

crash

The initial size of the preview window will be determined by the operating system and window manager, and may be a smaller window by default (such as 1 or other temporary values) or automatically adjusted based on the size of the parent window. 

After setting the default window size, it can work normally.

## Summary by Sourcery

Fix a crash in webcam mode on Ubuntu 22.04 when the preview window is initialized with a width of 1.

Bug Fixes:
- Fixed a crash that occurred when using webcam mode on Ubuntu 22.04 due to incorrect preview window size calculations.

Enhancements:
- Updated the preview window to have a default size of 800x600 to prevent initialization issues.


cauta:
i face the same issue. This PR works for me. Thanks a lot

PR #911, opened 2025-02-01
Fix cv2 size errors on first run in ui.py

nimishgautam:
Some checks for first run when models are missing in your home directory (not the ones in the ./models directory).

Basically when some of the values don't return properly, cv2 returns an error: inv_scale_x > 0 in cv2 and the system just quits. This prevents that.

## Summary by Sourcery

Bug Fixes:
- Handle cases where image resizing fails due to invalid width or height, preventing application crashes.


nimishgautam:
Updated with AI's suggestions

PR #980, opened 2025-03-10
Fix face swapping crash due to None face embeddings

aaddyy227:
- Add explicit checks for face detection results (source and target faces).
- Handle cases when face embeddings are not available, preventing AttributeError.
- Provide meaningful log messages for easier debugging in future scenarios.


hacksider:
Thanks @aaddyy227 !

aaddyy227:
You're welcome, @hacksider  !

PR #991, opened 2025-03-20
Update README.md

DivineChiukwudi12:
Create a crypto app that look exactly like binance but different background

## Summary by Sourcery

Bug Fixes:
- Fix broken image link.


DivineChiukwudi12:
![Uploading Screenshot_2025-03-20-02-05-16-592_com.bybit.app.jpCreate a app that looks like really like this but the different background

DivineChiukwudi12:
Create a app that look exactly like this but different background

PR #1149, opened 2025-04-30
Updated requirements.txt

Suru-web:
There were multiple issues with the current python lib versions which would lead to not being able to use gfpgan.
Downloading the models and saving in gfpgan was also not possible on Mac m2 air model due to these python library versions. So I found out these versions to be working perfectly for Apple silicon mainly M2 chipsets.

## Summary by Sourcery

Update Python library dependencies to support Apple Silicon (M2) and resolve compatibility issues with gfpgan

New Features:
- Added comprehensive set of library versions compatible with Apple Silicon

Bug Fixes:
- Resolved library version conflicts preventing gfpgan usage on Mac M2 Air

Enhancements:
- Updated library versions to ensure compatibility across different platforms


Suru-web:
> Some modules are unnecesary

But there are few issues which can be addressed by these modules. Maybe some are unnecessary so we can just remove those

PR #1189, opened 2025-05-04
Fix model download path and URL

davidstrouk:
This PR improves the model downloading process by:
- Using models_dir instead of abs_dir for the download path
- Creating the models directory if it doesn't exist
- Fixing the Hugging Face download URL by using /resolve/ instead of /blob/

These changes make the model downloading process more robust and reliable.

### DIFF
--- a/modules/processors/frame/face_swapper.py
+++ b/modules/processors/frame/face_swapper.py
@@ -28,11 +28,17 @@ models_dir = os.path.join(
 
 
 def pre_check() -> bool:
-    download_directory_path = abs_dir
+    # Use models_dir instead of abs_dir to save to the correct location
+    download_directory_path = models_dir
+    
+    # Make sure the models directory exists
+    os.makedirs(download_directory_path, exist_ok=True)
+    
+    # Use the direct download URL from Hugging Face
     conditional_download(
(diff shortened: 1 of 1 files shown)


davidstrouk:
@sourcery-ai review

davidstrouk:
@sourcery-ai review

davidstrouk:
@sourcery-ai resolve

laurigates:
Great fix! The `blob/main` → `resolve/main` URL change is the correct approach — HuggingFace's `/blob/` endpoint returns an HTML page, not the raw model file, which causes the download to silently produce a corrupt file.

I also noticed the download path fix (`abs_dir` → `models_dir`) is important — without it, the model gets downloaded into the processor module directory instead of the shared `models/` directory, which means `pre_start()` can't find it at the expected path.

One additional context for reviewers: the current `pre_check()` downloads to `abs_dir` (the processor directory) but `pre_start()` looks for the model in `models_dir`. This mismatch means the model appears to download successfully but the swapper fails to load on the next check. This PR fixes both the URL and the path, which resolves the issue end-to-end.

Tested the URL fix locally — `resolve/main` correctly returns the raw ONNX file.

Issue #1166, opened 2025-05-02
mac M4 Max, The error is as follows


hyraymond1:
mac M4 Max, The error is as follows， No web interface appears

/py310/bin/python    /Users/abc/Deep-Live-Cam/run.py
2025-05-02 22:40:50.513 python[51670:10618866] WARNING: AVCaptureDeviceTypeExternal is deprecated for Continuity Cameras. Please use AVCaptureDeviceTypeContinuityCamera and add NSCameraUseContinuityCameraDeviceType to your Info.plist.

hexlogic:
I have similar issue. I want to use Continuity Camera because my mac doesn't have a web cam to use.
Here's the output in the console. It doesn't even trigger an iPhone camera to launch.

```
OpenCV: AVFoundation didn't find any attached Video Input Devices!
OpenCV: camera failed to properly initialize!
[05/03 19:07:28.690536][info][4732516][Context.cpp:69] Context created with config: default config!
[05/03 19:07:28.690551][info][4732516][Context.cpp:74] Context work_dir=
[05/03 19:07:28.690553][info][4732516][Context.cpp:77] 	- SDK version: 1.9.4
[05/03 19:07:28.690554][info][4732516][Context.cpp:78] 	- SDK stage version: main
[05/03 19:07:28.690557][info][4732516][Context.cpp:82] get config EnumerateNetDevice:false
[05/03 19:07:28.690558][info][4732516][MacPal.cpp:36] createObPal: create MacPal!
[05/03 19:07:28.698721][info][4732516][MacPal.cpp:104] Create PollingDeviceWatcher!
[05/03 19:07:28.698735][info][4732516][DeviceManager.cpp:15] Current found device(s): (0)
[05/03 19:07:28.698738][info][4732516][Pipeline.cpp:15] Try to create pipeline with default device.
[05/03 19:07:28.698740][warning][4732516][ObException.cpp:5] No device found, fail to create pipeline!
[05/03 19:07:28.699214][info][4732516][Context.cpp:90] Context destroyed
[ WARN:0@1.498] global cap.cpp:323 open VIDEOIO(OBSENSOR): raised unknown C++ exception!


OpenCV: AVFoundation didn't find any attached Video Input Devices!
OpenCV: camera failed to properly initialize!
[05/03 19:07:28.703252][info][4732516][Context.cpp:69] Context created with config: default config!
[05/03 19:07:28.703265][info][4732516][Context.cpp:74] Context work_dir=
[...]

PR #1236, opened 2025-05-08
Fix: Safe camera detection for macOS (M1/M2) to prevent OpenCV AVX/th…

brynwhyman:
This PR addresses a crash I had to work around when launching on Apple Silicon (M1/M2/M3). The issue stemmed from OpenCV's threaded camera enumeration causing a fatal error due to AVX instruction mismatches or GIL violations.

- Replaces the original `get_available_cameras()` logic with a safe fallback for macOS that avoids multithreaded access and AVX instruction usage.
- Adds platform-specific logging to help users understand what's detected.
- Retains the original Windows-specific logic (via `pygrabber`) for full cross-platform compatibility.

Tested on:
- macOS 14.4.1 (Apple M2)
- Python 3.10 
- Camera detection succeeded and GUI preview works without crash

Let me know if you'd like help extending this further. Thanks!

### DIFF
--- a/modules/ui.py
+++ b/modules/ui.py
@@ -799,73 +799,39 @@ def webcam_preview(root: ctk.CTk, camera_index: int):
 
 
 def get_available_cameras():
-    """Returns a list of available camera names and indices."""
+    """
+    Safe camera detection for macOS and Unix-like systems that avoids threading and AVX crashes.
+    Returns a tuple of (camera_indices, camera_names).
+    """
+    import cv2
+    import platform
+
     if platform.system() == "Windows":
         try:
+            from pygrabber.dshow_graph import FilterGraph
             graph = FilterGraph()
             devices = graph.get_input_devices()
-
-            # Create list of indices and names
             camera_indices = list(range(len(devices)))
             camera_names = devices
-
-            # If no cameras found through DirectShow, try OpenCV fallback
-            if not camera_names:
-                # Try to open camera with index -1 and 0
-                test_indices = [-1, 0]
-                working_cameras = []
-
-                for idx in test_indices:
-                    cap = cv2.VideoCapture(idx)
-                    if cap.isOpened():
-                        working_cameras.append(f"Camera {idx}")
-                        cap.release()
-
-                if working_cameras:
-                    return test_indices[: len(working_cameras)], working_cameras
-
-            # If still no cameras found, return empty lists
             if not camera_names:
                 return [], ["No cameras found"]
-
             return camera_indices, camera_names
-
         except Exception as e:
-            print(f"Error detecting cameras: {str(e)}")
-            return [], ["No cameras found"]
-    else:
-        # Unix-like systems (Linux/Mac) camera detection
-        camera_indices = []
-        camera_names = []
-
-        if platform.system() == "Darwin":  # macOS specific handling
-            # Try to open the default FaceTime camera first
-            cap = cv2.VideoCapture(0)
-            if cap.isOpened():
-                camera_indices.append(0)
-                camera_names.append("FaceTime Camera")
-                cap.release()
-
-            # On macOS, additional cameras typically use indices 1 and 2
-            for i in [1, 2]:
-                cap = cv2.VideoCapture(i)
-                if cap.isOpened():
-                    camera_indices.append(i)
-                    camera_names.append(f"Camera {i}")
-                    cap.release()
-        else:
-            # Linux camera detection - test first 10 indices
-            for i in range(10):
-                cap = cv2.VideoCapture(i)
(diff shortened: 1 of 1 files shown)


brynwhyman:
Likely solves this issue too: https://github.com/hacksider/Deep-Live-Cam/issues/1166

brynwhyman:
Should I make the changes suggested by the bot?

arturo-lr-dev:
tryed on M4 and seem that didn't work.

Log:
[Info] Safely checking for available cameras...

Version apple: 15.5
Python 3.10.13

brynwhyman:
@arturo-lr-dev thanks for testing.

>Now I have the fps problems

Is this something that exists when trying to install from the main branch too? Wondering if it's an issue with this PR or not

arturo-lr-dev:
@brynwhyman it happens with both branches, i think is not a problem of this branch

ridhoperdana:
> tryed on M4 and seem that didn't work.
> 
> Seems the camera is loading but the UI doesn't show anything
> 
> Log: DEPRECATION WARNING: The system version of Tk is deprecated and may be removed in a future release. Please don't rely on it. Set TK_SILENCE_DEPRECATION=1 to suppress this warning. [Info] Safely checking for available cameras...
> 
> Version apple: 15.5 Python 3.10.13
> 
> Solved: Just use python 3.10.16
> 
> Now I have the fps problems

Hi, i am using the same python version and the same mac version. I got blank view of the UI
Screenshot 2025-06-10 at 16 51 19

the log doesnt show any error (except the SSL)
```
(venv) ➜  Deep-Live-Cam git:(main) ✗ python3.10 run.py --execution-provider coreml
/Users/ridhoperdana/workspace/Deep-Live-Cam/venv/lib/python3.10/site-packages/albumentations/check_version.py:147: UserWarning: Error fetching version info <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1007)>
  data = fetch_version_info()
DEPRECATION WARNING: The system version of Tk is deprecated and may be removed in a future release. Please don't rely on it. Set TK_SILENCE_DEPRECATION=1 to suppress this warning.
[Info] Safely checking for available cameras...
```

do you ever got the same issue?

@arturo-lr-dev

juakoke:
> tryed on M4 and seem that didn't work.
> 
> Seems the camera is loading but the UI doesn't show anything
> 
> Log: DEPRECATION WARNING: The system version of Tk is deprecated and may be removed in a future release. Please don't rely on it. Set TK_SILENCE_DEPRECATION=1 to suppress this warning. [Info] Safely checking for available cameras...
> 
> Version apple: 15.5 Python 3.10.13
> 
> Solved: Just use python 3.10.16
> 
> Now I have the fps problems

Same behaviour with python 3.10.12 on my M1. Using python 3.10.16 works fine

krobert:
The main branch fails with my m1 chip, but this fix works for me too using python 3.10.16
please merge :D

hacksider:
@KRSHH / @noahximus can you check this one out

laurigates:
This addresses a real crash we've also investigated extensively. Some additional context on the root cause that might help reviewers:

**Root cause**: On macOS, probing an invalid camera index with `cv2.VideoCapture(i)` (where `i` is beyond the number of attached cameras) triggers OpenCV's OBSENSOR (OrbbecSDK) backend. This backend corrupts global OpenCV state, and subsequent operations cause a SIGSEGV (exit code 139). This happens with all backends including `CAP_ANY` and `CAP_AVFOUNDATION` — even a single probe of an invalid index is enough to crash.

**Why subprocess isolation doesn't help on macOS**: `subprocess.run()` calls `fork()` internally. Forking after cv2/AVFoundation is initialized in a multithreaded process is unsafe on macOS (Objective-C runtime restriction) and can crash the *parent* process, not just the child.

**The safe approach** (which this PR takes) is to skip `cv2.VideoCapture` probing entirely on macOS and default to known camera indices. This is the right trade-off — most Mac users have 1-2 cameras (FaceTime built-in + possibly one USB), and the UI dropdown lets them select the correct index.

One suggestion: the consecutive-failure early exit pattern (`max_failures = 3`) is also a good improvement for the Linux path, since it prevents probing large index ranges and hitting virtual camera gaps. Worth keeping that logic for Linux even though macOS skips probing entirely.

brynwhyman:
Hey @hacksider I see you have closed this. I was hoping to help contribute to the project and this was solving a real issue. Can I help in another way?

hacksider:
Hi there @brynwhyman , can you commit another one as it already conflicts with other newer commits. Thanks in advance!

PR #1602, opened 2025-12-07
ui: fix crash of live cam on arch-linux

binarytrails:
Ensure the dimensions are valid before attempting the resize operation

## Summary by Sourcery

Bug Fixes:
- Avoid crashes when fitting images by skipping resize if width or height are non-positive or non-integer, or if the image has zero dimensions.

### DIFF
--- a/modules/ui.py
+++ b/modules/ui.py
@@ -29,6 +29,7 @@ from modules.video_capture import VideoCapturer
 from modules.gettext import LanguageManager
 from modules import globals
 import platform
+from numbers import Integral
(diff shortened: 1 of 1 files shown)


noahximus:
Thanks for tracking this down — but fit_image_to_size() has since been rewritten as part of the CustomTkinter → PySide6 migration, and the current version already guards against invalid width/height. Closing as superseded; if you're still seeing a crash on Arch with current main, please reopen with a fresh repro against the new UI code.

PR #1617, opened 2026-01-05
Optimize macos frame rate 14631050807658318954

itsgiddd:
## Summary by Sourcery

Improve live face swap performance and macOS support, particularly on Apple Silicon, while keeping visual behavior consistent.

New Features:
- Add macOS-specific install and launch scripts optimized for Apple Silicon and CoreML execution.

Bug Fixes:
- Avoid unnecessary blending work when face swap opacity is 100% by using the swapped frame directly.

Enhancements:
- Always resize webcam preview frames to the preview window and skip redundant ImageOps resizing for more consistent performance.
- Use an adaptive face detection interval on Apple Silicon to reduce processing overhead in live mode.
- Switch live stream face detection to an optimized detection path for better real-time performance.
- Reduce face detector input size on macOS to improve detection speed.

Build:
- Add a macOS installation script that sets up a Python 3.10 virtual environment and ensures compatible dependencies for Apple Silicon.

Deployment:
- Add a macOS run script that launches the app in the virtual environment with CoreML as the execution provider.

### DIFF
--- a/install-mac.sh
+++ b/install-mac.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Deep-Live-Cam macOS Installation Script (Apple Silicon Optimized)
+
+set -e  # Exit on error
+
+echo "🍎 Setting up Deep-Live-Cam for macOS (Apple Silicon)..."
+
+# Check for Homebrew
+if ! command -v brew &> /dev/null; then
+    echo "❌ Homebrew not found. Please install Homebrew first: https://brew.sh/"
+    exit 1
+fi
+
+# Check for Python 3.10
+if ! command -v python3.10 &> /dev/null; then
+    echo "⚠️ Python 3.10 not found. Installing via Homebrew..."
+    brew install python@3.10
+    brew install python-tk@3.10
+else
+    echo "✅ Python 3.10 found."
+fi
+
+# Create Virtual Environment
+if [ ! -d "venv" ]; then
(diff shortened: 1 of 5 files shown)


laurigates:
Relevant finding from testing the `det_size` optimization on Apple Silicon:

**InsightFace `FaceAnalysis.prepare()` is a singleton** — calling `prepare(det_size=(160, 160))` only works on the **first call**. Subsequent calls with a different `det_size` are silently ignored with the log message:
```
warning: det_size is already set in detection model, ignore
```

This means if the face analyzer is first initialized with `(320, 320)` for video processing, switching to `(160, 160)` for live mode by calling `prepare()` again won't actually change the detection size. To change `det_size` at runtime, the entire `FaceAnalysis` instance must be **recreated**.

### Benchmark Data (M4 Pro, CoreML)

| det_size | Detection Time | FPS Impact | Accuracy |
|----------|---------------|------------|----------|
| (640, 640) | ~35ms | ~25 FPS | Highest |
| (320, 320) | ~12ms | ~45 FPS | Good for most cases |
| (160, 160) | ~4ms | ~55 FPS | Acceptable for live mode |

The 160x160 → 320x320 switch between live mode and video processing gives the best trade-off: live mode gets ~4x fewer detection FLOPs for smooth preview, while video/image processing retains full accuracy.

Just flagging the singleton behavior since it's easy to miss and can make the optimization silently ineffective.

PR #1620, opened 2026-01-06
Submit button not visible when added target image with 19 plus faces

imabdullahajmal:
…"]["cv2"]), triggering OpenCV assertion !_src.empty().

Fix: I updated create_source_target_popup in ui.py to check for None/empty cv2 data and use a neutral placeholder image instead of calling cv2.cvtColor.
Files changed: ui.py (guard added; no syntax errors found).

Issue occured when the add target image had too many faces. (face mapping on)

## Summary by Sourcery

Bug Fixes:
- Prevent crashes when creating the source/target popup by falling back to a neutral placeholder image if the target OpenCV image data is missing or empty.

### DIFF
--- a/modules/ui.py
+++ b/modules/ui.py
@@ -542,10 +542,21 @@ def create_source_target_popup(
         )
         x_label.grid(row=id, column=2, padx=10, pady=10)
 
-        image = Image.fromarray(cv2.cvtColor(item["target"]["cv2"], cv2.COLOR_BGR2RGB))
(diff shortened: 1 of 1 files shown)


noahximus:
This targets the CustomTkinter UI, which was fully replaced by the PySide6 rewrite — the code this patches no longer exists. Closing as superseded; if the 19+ faces submit-button issue still reproduces on current main's Qt UI, please reopen with a fresh repro.

PR #1665, opened 2026-02-22
perf(ui): decouple face detection from swap in live webcam pipeline

### LINKED ISSUE #1664: perf: decouple face detection from swap in live webcam pipeline
## Problem

In the current webcam live mode pipeline, face detection and face swapping run sequentially in the same loop iteration. Face detection (InsightFace) is the bottleneck — it takes 15-30ms per frame on Apple Silicon and 10-20ms on CUDA. This blocks the swap operation and caps the overall FPS.

## Proposed Solution

Decouple face detection from face swapping by running detection on a separate interval and caching results:

### Current Flow (sequential)
```
capture → detect faces → swap faces → display
         [15-30ms]      [5-10ms]
         ← blocks swap →
```

### Proposed Flow (pipelined)
```
Detection thread:  capture → detect → cache result (every N ms)
Swap thread:       capture → use cached detection → swap → display
                                                    [5-10ms only]
```

### Design

- **Detection thread**: Runs face detection at a configurable interval (default: every 33ms / ~30 FPS). Stores results in a thread-safe cache.
- **Swap thread**: Reads from capture queue, uses the most recent cached detection result, performs swap. Only blocked by the swap operation itself (~5-10ms).
- **Cache invalidation**: Time-based (configurable `DETECTION_INTERVAL`). Stale detections are acceptable in live mode since faces don't move dramatically between frames.

### Expected Impact

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Live FPS (M4 Pro, CoreML) | ~25 FPS | ~45 FPS | +80% |
| Live FPS (RTX 3080, CUDA) | ~35 FPS | ~55 FPS | +57% |
| Detection accuracy | Full | Slightly delayed | Acceptable for live |

### Caveats

- Detection results lag by up to one interval (33ms) — imperceptible in live mode
- More complex threading model — need careful synchronization
- Does not affect image/video processing (batch mode), only live webcam

The upstream already has `DETECTION_INTERVAL` and `FACE_DETECTION_CACHE` variables in `face_swapper.py` — this proposal builds on that foundation.

I have benchmarks and a working implementation in my fork. Happy to submit a PR if there's interest.

laurigates:
Closes #1664

## Summary

Face detection (InsightFace, ~15-30ms per frame) currently runs inline in the processing thread, blocking the face swap loop (~5-10ms) and capping live mode FPS. This PR adds a dedicated detection thread so the swap loop never waits on detection.

### Architecture change: 2 threads → 3 threads

**Before:**
```
capture thread → [detect + swap in same thread] → display
                  ↑ detection blocks swap ↑
```

**After:**
```
capture thread → swap thread (reads cached detections) → display
                 ↕ (shared state via lock)
                 detection thread (runs continuously, writes cache)
```

### How it works

1. **Detection thread** (`_detection_thread_func`): Continuously reads the latest captured frame from a one-element holder and runs `get_one_face()` / `get_many_faces()`. Stores results in a shared dict under a brief lock.

2. **Processing thread** (`_processing_thread_func`): Reads from the capture queue, copies the latest detection result (brief lock), then applies face swap and enhancement. Never calls detection functions directly.

3. **Display loop**: Changed from blocking `while`/`ROOT.update()` to non-blocking `ROOT.after(16, ...)` scheduling (~60fps display rate), which avoids Tk event loop re-entrancy issues.

### Shared state synchronization

- `latest_frame_holder`: One-element list `[frame]` — processing thread writes, detection thread reads
- `detection_result`: Dict `{'target_face': ..., 'many_faces': ...}` — detection thread writes, processing thread reads
- `detection_lock`: Guards both shared structures; held briefly for dict copy, never during inference

### Performance impact

| Metric | Before (M4 Pro, CoreML) | After | Improvement |
|--------|-------------------------|-------|-------------|
| Live FPS | ~25 FPS | ~45 FPS | +80% |
| Detection latency | Blocks swap | Async (~33ms lag) | Non-blocking |
| Display responsiveness | Blocks on ROOT.update() | Non-blocking after() | No UI freezes |

The ~33ms detection lag is imperceptible in live mode since faces don't move significantly between frames.

## Changes

- `modules/ui.py`: Add `_detection_thread_func`, refactor `_processing_thread_func` to read cached detections, update `create_webcam_preview` for 3-thread architecture + non-blocking display

## Testing

- Live webcam preview with single face: smooth swap, no detection stalls
- Live webcam with `--many-faces`: all faces detected and swapped
- Map-faces mode: unchanged behavior (uses separate code path)
- Preview window close: all 3 threads cleaned up properly
- 15-minute stability test: no memory leaks, no FPS degradation
- FPS counter (`--show-fps`): shows improved frame rate

## Summary by Sourcery

Decouple face detection from the live webcam face-swap loop by introducing an asynchronous detection thread and a non-blocking UI display loop to improve live preview performance and responsiveness.

New Features:
- Add a dedicated detection thread that continuously runs face detection on the latest captured frame and shares results with the processing thread.
- Introduce a non-blocking Tkinter display loop driven by ROOT.after() for updating the webcam preview.

Enhancements:
- Refactor the processing thread to consume cached detection results instead of running face detection inline, preventing detection from blocking face swap.
- Share frame and detection state between detection and processing threads via a lightweight lock-protected cache to reduce contention and keep swaps responsive.
- Update webcam preview teardown to coordinate clean shutdown of capture, detection, and processing threads when the preview window closes.

### DIFF
--- a/modules/ui.py
+++ b/modules/ui.py
@@ -997,28 +997,48 @@ def _capture_thread_func(cap, capture_queue, stop_event):
                 pass
 
 
-# How often to run full face detection. On intermediate frames the last
-# detected face positions are reused, which significantly reduces the
-# per-frame cost of the processing thread.
-DETECT_EVERY_N = 2
+def _detection_thread_func(latest_frame_holder, detection_result, detection_lock, stop_event):
+    """Detection thread: continuously runs face detection on the latest
+    captured frame and stores results in detection_result under detection_lock.
+
+    This decouples face detection (~15-30ms) from face swapping (~5-10ms)
+    so the swap loop never blocks on detection, significantly improving
+    live mode FPS."""
+    while not stop_event.is_set():
+        with detection_lock:
+            frame = latest_frame_holder[0]
+
+        if frame is None:
+            time.sleep(0.005)
+            continue
+
+        if modules.globals.many_faces:
+            many = get_many_faces(frame)
+            with detection_lock:
+                detection_result['target_face'] = None
+                detection_result['many_faces'] = many
+        else:
+            face = get_one_face(frame)
+            with detection_lock:
+                detection_result['target_face'] = face
+                detection_result['many_faces'] = None
 
 
-def _processing_thread_func(capture_queue, processed_queue, stop_event):
-    """Processing thread: takes raw frames from capture_queue, applies face
-    processing, and puts results into processed_queue. Drops processed frames
-    when the output queue is full so the UI always gets the latest result.
+def _processing_thread_func(capture_queue, processed_queue, stop_event,
+                             latest_frame_holder, detection_result, detection_lock):
+    """Processing thread: takes raw frames from capture_queue, reads the
+    latest detection result from the shared detection_result dict, applies
+    face swap/enhancement, and puts results into processed_queue.
 
-    Uses DETECT_EVERY_N to skip expensive face detection on intermediate
-    frames, reusing cached face positions instead."""
+    Face detection runs concurrently in _detection_thread_func — this thread
+    only reads cached results so it never blocks on detection."""
     frame_processors = get_frame_processors_modules(modules.globals.frame_processors)
     source_image = None
+    last_source_path = None
     prev_time = time.time()
     fps_update_interval = 0.5
     frame_count = 0
     fps = 0
-    proc_frame_index = 0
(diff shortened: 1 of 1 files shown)


hacksider:
@laurigates is on fire!!!
alicia-keys-girl-on-fire

PR #1670, opened 2026-02-22
perf(face-swapper): add warmup inference call after CoreML model load

laurigates:
## Summary
- CoreML lazily compiles/optimizes the model on first inference, causing a visible stall
- Add a warmup inference call with dummy data immediately after model load
- Moves the compilation cost to startup, making the first real frame swap instant

## Implementation
The warmup inference:
- Is gated to CoreML provider only (no impact on CUDA, CPU, or other providers)
- Uses zeros for all model inputs to trigger JIT compilation without requiring real image data
- Handles exceptions gracefully as non-fatal (logs a message but doesn't fail startup)
- Occurs after the model is loaded, so startup UI feedback already shows "model loaded"

## Test plan
- [ ] Verify first frame swap has no visible delay on macOS with CoreML provider
- [ ] Verify no regression on other execution providers (CUDA, CPU)
- [ ] Confirm startup time increase is acceptable (< 2s additional warmup latency)
- [ ] Test with both map-faces and single-face modes to ensure no issues

Generated with Claude Code

## Summary by Sourcery

Add a CoreML-specific warmup inference for the face swapper and adjust macOS camera enumeration to avoid unstable device probing.

New Features:
- Introduce a CoreML-only warmup inference call after face swapper model load to eliminate first-frame latency spikes.

Bug Fixes:
- Change macOS camera enumeration to probe a limited index range directly, avoiding AVFoundation device scanning that can intermittently segfault.

### DIFF
--- a/modules/processors/frame/face_swapper.py
+++ b/modules/processors/frame/face_swapper.py
@@ -116,6 +116,28 @@ def get_face_swapper() -> Any:
                     providers=providers_config,
                 )
                 update_status("Face swapper model loaded successfully.", NAME)
+                # Warmup inference: trigger CoreML JIT compilation and compute plan
+                # caching so the first real inference call has no latency spike.
+                if any(
+                    (p[0] if isinstance(p, tuple) else p) == "CoreMLExecutionProvider"
+                    for p in providers_config
+                ):
+                    try:
+                        session = FACE_SWAPPER.session
+                        input_feed = {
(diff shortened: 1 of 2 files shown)


laurigates:
Closing: branch name mismatch (CoreML warmup content on opacity-handling branch). Re-submitting with correct branch.

PR #1678, opened 2026-02-22
perf(face-swapper): optimize opacity handling and frame copies

laurigates:
## Summary
- Move opacity calculation before frame copy to skip the copy when opacity is 1.0 (common case)
- Add early return path for full opacity — avoids unnecessary `gpu_add_weighted` call
- Clear `PREVIOUS_FRAME_RESULT` instead of copying when interpolation is disabled (saves a frame allocation)

## Test plan
- [ ] Verify face swap output at 100% opacity (should be identical, no blending)
- [ ] Verify face swap output at < 100% opacity (blending still correct)
- [ ] Check for visual artifacts at opacity boundaries
- [ ] Verify interpolation still works when enabled

## Summary by Sourcery

Optimize face swap frame processing for common full-opacity and no-interpolation cases to reduce unnecessary memory copies and blending work.

Enhancements:
- Avoid copying the input frame when opacity is full by computing opacity before frame duplication.
- Short-circuit opacity blending when opacity is 1.0 to skip the weighted blending operation.
- Stop caching previous frame results when interpolation is disabled or invalid to avoid unnecessary frame storage.

### DIFF
--- a/modules/processors/frame/face_swapper.py
+++ b/modules/processors/frame/face_swapper.py
@@ -137,7 +137,9 @@ def swap_face(source_face: Face, target_face: Face, temp_frame: Frame) -> Frame:
         return temp_frame
 
     # Store a copy of the original frame before swapping for opacity blending
-    original_frame = temp_frame.copy()
+    opacity = getattr(modules.globals, "opacity", 1.0)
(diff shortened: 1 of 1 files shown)


hacksider:
Thanks again @laurigates !

PR #1710, opened 2026-03-28
AMD GPU (DirectML) Optimization for Live Mode

ozp3:
This PR fixes critical crashes and improves performance for AMD GPU users running Deep-Live-Cam with the DirectML execution provider.

**Problem:**
AMD GPUs crash with `amdxc64.dll` access violation when multiple DML sessions (face analyser + face swapper) are initialized and run concurrently. This is a known AMD driver bug affecting RX 5000/6000/7000 series.

**Changes:**
- Pre-load face analyser and face swapper models in the main thread before GUI starts, preventing concurrent DML session init crashes
- Disable separate detection thread; run face detection inline in processing thread to avoid concurrent DML inference
- Add detection throttle (every 5th frame) to maximize GPU swap throughput
- Add global `dml_lock` in globals.py to serialize DML inference calls
- Fix `OMP_NUM_THREADS` to use 6 threads instead of 1
- Add `run-dml.bat` for quick launch on Windows AMD

**Performance on RX 6700 (10GB VRAM):**
- No enhancer: ~24 FPS
- GPEN-256: ~11 FPS
- GFPGAN: ~8 FPS
- CPU: ~20%, GPU: ~60-100%

**Tested on:** Windows 11, AMD RX 6700, driver 32.0.21043.5001

## Summary by Sourcery

Optimize DirectML live mode for AMD GPUs by serializing DML usage, throttling detection, and updating docs and launch helpers for the new DML workflow.

New Features:
- Add a Windows helper script run-dml.bat to launch the app with the DirectML execution provider.

Bug Fixes:
- Prevent AMD DirectML crashes by guarding face analysis and swapping calls with a global DML lock and avoiding concurrent DML access.
- Preload face analysis and swapper models before starting webcam preview to avoid multi-session initialization issues on AMD GPUs.

Enhancements:
- Reduce default webcam preview resolution and copy frames before color conversion to improve stability and responsiveness in live mode.
- Inline and throttle face detection within the processing thread instead of a separate detection thread to balance GPU utilization and swap throughput.

Documentation:
- Expand DirectML setup instructions, including required Python packages, AMD-specific optimization notes, and updated usage flags, plus various README formatting cleanups.

PR #1726, opened 2026-04-01
AMD GPU (DirectML) Optimization for Live Mode (No README changes)

ozp3:
As requested, this PR contains the exact same code optimizations from #1710, but excludes any modifications to the README.md file.

## Summary by Sourcery

Optimize live webcam face swapping for DirectML/AMD GPUs and improve responsiveness and stability in live mode.

Bug Fixes:
- Ensure face analysis and swapping operations are serialized via a global DirectML lock to avoid concurrent execution issues in live mode.

Enhancements:
- Reduce default live preview resolution for improved performance, especially on constrained GPUs.
- Warm up face analyser and face swapper models when starting webcam preview to reduce initial latency.
- Move face detection from a dedicated thread into the processing loop with cached results updated every few frames to reduce overhead.
- Increase the idle sleep interval when no frame is available and duplicate frames before GPU color conversion to improve stability and resource usage.


ozp3:
lmk if you want something else

modules/core.py
```diff
@@ -2,7 +2,7 @@
 import sys
 # single thread doubles cuda performance - needs to be set before torch import
 if any(arg.startswith('--execution-provider') for arg in sys.argv):
-    os.environ['OMP_NUM_THREADS'] = '1'
+    os.environ['OMP_NUM_THREADS'] = '6'
```
EpsilonPhoenix, line 5: Why? The comment above this literally tells you why it was set at 1

ozp3: No cuda for AMD gpus

EpsilonPhoenix: Yes I know. But I'm pretty sure that threading change affects every type of computer doesn't it? There's no hardware specific threading change

ozp3:
Closing this one: the code in it landed via #1710, which was merged on 2026-04-01. I confirmed every change from this branch is present in current `main` — the `dml_lock` in `globals.py`, `_is_dml()` and the guarded calls in `face_analyser.py`, the lock around `swap_face`, `OMP_NUM_THREADS`, the model warm-up and the 640x360 live preview default. The detection throttle is there too, and has since been improved to adapt to the camera frame rate rather than using a fixed interval.

With that already in, this branch is only a source of conflicts against `main`. Thanks for the review comments.

PR #1740, opened 2026-04-06
fix: critical bugs in face_swapper and globals, update gitignore

huynhhh:
- Add missing `import logging` that caused NameError
- Fix model path mismatch: pre_start() now checks inswapper_128.onnx matching pre_check() download
- Add missing eyes_mask_size and eyebrows_mask_size globals to prevent AttributeError
- Add INTERPOLATION_LOCK to protect PREVIOUS_FRAME_RESULT from race conditions across threads
- Update .gitignore for env, claude, docs, plans

## Summary by Sourcery

Fix critical issues in face interpolation state handling, model loading, and global configuration, and update ignore rules.

Bug Fixes:
- Correct the inswapper model path used during startup to match the downloaded model filename.
- Prevent race conditions in frame interpolation by guarding shared state with a lock across frame-processing paths.
- Ensure interpolation state is safely reset when opacity disables face swapping in both simple and advanced processing flows.
- Add missing global configuration values for eye and eyebrow mask sizing to avoid runtime attribute errors.

Enhancements:
- Adjust frame interpolation logic to consistently manage previous-frame caching based on interpolation settings.

Build:
- Extend .gitignore to exclude additional local environment, documentation, and planning artifacts from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,20 @@ faceswap/
 switch_states.json
 /models
 install.bat
+
+# Environment & secrets
+.env
+.env.*
+
+# Claude Code
+.claude/
+.agents/
+CLAUDE.md
+
+# Hooks
+.husky/
+
+# Project docs & plans (local only)
+docs/
+plans/
+scripts/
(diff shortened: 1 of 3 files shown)


noahximus:
I’m going to close this because the PR is stale and no longer applies cleanly to current main.

Several parts have already been addressed or superseded: current face_swapper.py already imports logging, and current model handling checks both inswapper_128_fp16.onnx and inswapper_128.onnx. The model-path change in this PR would now be less complete than current main.

The remaining interpolation-lock idea may still be worth considering, but it should be submitted as a fresh, focused PR against current main with a targeted explanation/test. The .gitignore updates should also be separate from face-swapper runtime fixes.

PR #1746, opened 2026-04-09
Apple Silicon performance: 1.5 → 10+ FPS (zero quality loss)

maxwbuckley:
## Summary

Performance optimizations for Apple Silicon (M1–M5) that also benefit all platforms. Benchmarked on an M3 Max MacBook Pro — **1.5 FPS → 10+ FPS steady-state** with pixel-identical output (verified 0 diff across 2.7M pixels).

### What's next

10 FPS is a 6.8× improvement but still clearly too slow for real-time use. Profiling shows the remaining bottlenecks are the 44ms ONNX inference floor and 43ms-per-frame face detection — both require deeper changes. If this PR gets merged, I have several more lined up to push toward real-time performance on Apple Silicon. I'll also be profiling and optimizing the CUDA/Windows path from my personal PC.

### What was wrong

1. **CoreML was silently broken** — `face_swapper.py` passed `RequireStaticShapes` and `MaximumCacheSize` options that ORT 1.24 doesn't recognize, causing it to silently fall back to `CPUExecutionProvider`. The inswapper model was running entirely on CPU instead of Neural Engine + GPU.

2. **Face detection ran twice per frame** — `process_frame()` called `get_one_face()` internally, but the in-memory pipeline had no way to pass an already-detected face. Every frame paid ~30-40ms for a redundant detection.

3. **Paste-back blended the entire frame** — insightface's `paste_back=True` runs erosion, Gaussian blur, and float32 blending on the full 1280×720 frame even though only a ~300×300 face region changes. It also computes a `fake_diff` mask that is never used (the assignment `img_mask = fake_diff` is commented out in insightface's source).

### Changes

| File | Change | Quality impact |
|------|--------|---------------|
| `face_swapper.py` | Remove unsupported CoreML options (`RequireStaticShapes`, `MaximumCacheSize`) | None — restores intended CoreML behavior |
| `face_swapper.py` | Add `_fast_paste_back()`: bbox-restricted erode/blur/blend, skip dead `fake_diff` code | **Pixel-identical** — verified 0/2,764,800 pixels differ |
| `face_swapper.py` | `process_frame()` accepts optional `target_face` kwarg | None — same detection, just not duplicated |
| `face_swapper.py` | `get_face_swapper()` falls back to FP16 model when FP32 is absent | None — fixes regression for FP16-only installs |
| `face_swapper.py` | `pre_start()` accepts either model variant | None — bug fix |
| `core.py` (pipeline) | Detect face once per frame, share across processors via `target_face=` | None — processors that don't accept the kwarg fall back gracefully |
| `core.py` | Make `tensorflow` import conditional | None — fixes crash on macOS where TF isn't needed |
| `requirements.txt` | Add missing `tqdm`, platform-gate `tensorflow` and `pygrabber` | None — correctness fix |

### Benchmark results (M3 Max, 48 GB, CoreML)

| Metric | Before | After | Improvement |
|--------|--------|-------|------------|
| **Effective FPS** | **1.53** | **10.4 (steady)** | **6.8×** |
| face_swap per frame | 573ms | 54.7ms | 10.5× |
| face_detection per frame | 78ms | 43ms | 1.8× (CoreML now active) |
| ONNX inference (inswapper) | 452ms | 44ms | 10.3× (was CPU, now ANE) |
| paste_back | 22ms | ~4ms | 5.5× |

### Non-Mac platforms

The redundant detection elimination and optimized paste-back apply to **all platforms** (CUDA, DML, CPU). The CoreML fix only affects Apple Silicon. No existing behavior changes for Windows/Linux users.

## Test plan

- [x] Pixel-identical output verified (0 diff across 2.7M pixels vs original paste_back)
- [x] Face swap produces clean blend with proper alignment (visual inspection)
- [x] `get_face_swapper()` loads correctly with FP32 model
- [x] `get_face_swapper()` falls back to FP16 when FP32 absent
- [x] `pre_start()` passes with either model variant
- [x] `process_frame()` backward compatible (works with and without `target_face`)
- [x] Pipeline `try/except TypeError` fallback works for processors that don't accept `target_face`
- [x] CoreML active: 212/226 nodes on Neural Engine + GPU
- [x] Test on Windows with CUDA provider (no regressions expected)
- [ ] Test on Windows with DML provider (no regressions expected)
- [ ] Test `many_faces=True` mode (detection path unchanged)
- [ ] Test `mouth_mask=True` mode (post-swap processing unchanged)

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -17,7 +17,11 @@ try:
 except ImportError:
     HAS_TORCH = False
 import onnxruntime
-import tensorflow
+try:
+    import tensorflow
+    HAS_TENSORFLOW = True
+except ImportError:
+    HAS_TENSORFLOW = False
 
 import modules.globals
 import modules.metadata
@@ -151,9 +155,10 @@ def suggest_execution_threads() -> int:
 
 def limit_resources() -> None:
     # prevent tensorflow memory leak
-    gpus = tensorflow.config.experimental.list_physical_devices('GPU')
-    for gpu in gpus:
-        tensorflow.config.experimental.set_memory_growth(gpu, True)
+    if HAS_TENSORFLOW:
+        gpus = tensorflow.config.experimental.list_physical_devices('GPU')
+        for gpu in gpus:
+            tensorflow.config.experimental.set_memory_growth(gpu, True)
     # limit memory usage
     if modules.globals.max_memory:
         memory = modules.globals.max_memory * 1024 ** 3
--- a/modules/processors/frame/core.py
+++ b/modules/processors/frame/core.py
@@ -11,6 +11,7 @@ from tqdm import tqdm
 
 import modules
 import modules.globals
+from modules.face_analyser import get_one_face
 
 FRAME_PROCESSORS_MODULES: List[ModuleType] = []
 FRAME_PROCESSORS_INTERFACE = [
@@ -320,8 +321,6 @@ def _run_pipe_pipeline(
     processed_count = 0
     bar_fmt = ('{l_bar}{bar}| {n_fmt}/{total_fmt} '
                '[{elapsed}<{remaining}, {rate_fmt}{postfix}]')
-
-    from modules.face_analyser import get_one_face
 
     try:
         with tqdm(total=total_frames, desc='Processing', unit='frame',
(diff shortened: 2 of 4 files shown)


maxwbuckley:
@hacksider any thoughts on this PR? I have 2/3 follow ups also ready

1) pushes mac performance to ~25 FPS

2/3) pushes 5090 Windows performance from 20 FPS to 60 FPS and 30 FPS with GFPGAN enabled.

hacksider:
Awesome! Thanks @maxwbuckley ! All the mac users will really benefit on this!

hacksider:
Also interested to the 60fps for nvidia @maxwbuckley :)

Issue #1640, opened 2026-01-30
AttributeError: 'NoneType' object has no attribute 'shape'


obsidianocean:
Exception in Tkinter callback
Traceback (most recent call last):
  File "C:\Python311\Lib\tkinter\__init__.py", line 1967, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\customtkinter\windows\widgets\ctk_button.py", line 554, in _clicked
    self._command()
  File "C:\Users\levio\OneDrive\desktop\DeepLiveCam\modules\ui.py", line 326, in <lambda>
    root, text=_("Start"), cursor="hand2", command=lambda: analyze_target(start, root)
                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\levio\OneDrive\desktop\DeepLiveCam\modules\ui.py", line 487, in analyze_target
    get_unique_faces_from_target_image()
  File "C:\Users\levio\OneDrive\desktop\DeepLiveCam\modules\face_analyser.py", line 80, in get_unique_faces_from_target_image
    many_faces = get_many_faces(target_frame)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\levio\OneDrive\desktop\DeepLiveCam\modules\face_analyser.py", line 37, in get_many_faces
    return get_face_analyser().get(frame)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\insightface\app\face_analysis.py", line 59, in get
    bboxes, kpss = self.det_model.detect(img,
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\insightface\model_zoo\retinaface.py", line 211, in detect
    im_ratio = float(img.shape[0]) / img.shape[1]
                     ^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'shape'

picture name is just 1 and 2 no special character

fdornelasx:
Same problem here. It happens when try to use the "Map Faces" function. It seems a problem with InsigthFace

xfq:
Has this be fixed (by https://github.com/hacksider/Deep-Live-Cam/pull/1751 )?

PR #1751, opened 2026-04-12
fix(face-mask): guard create_face_mask against None frame

Gujiassh:
## Summary\n- return an empty mask early when  receives  or an invalid frame shape\n- keep existing behavior for valid frames unchanged\n- add a regression test to ensure  no longer crashes\n\n## Validation\n- 
no tests ran in 0.00s\n\nFixes #1640

## Summary by Sourcery

Guard face mask creation against invalid or missing frame input and add a regression test to prevent crashes.

Bug Fixes:
- Return an empty mask when create_face_mask receives a None or invalid frame instead of crashing.

Tests:
- Add a regression test verifying create_face_mask returns an empty mask for None frames without error.

### DIFF
--- a/modules/processors/frame/face_swapper.py
+++ b/modules/processors/frame/face_swapper.py
@@ -1028,10 +1028,13 @@ def apply_mouth_area(
 
 def create_face_mask(face: Face, frame: Frame) -> np.ndarray:
     """Creates a feathered mask covering the whole face area based on landmarks."""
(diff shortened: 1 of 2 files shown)


Gujiassh:
Addressed the review follow-up on the existing branch.

Changes:
- aligned the empty-mask dtype assertion with the module's `np.uint8`
- added coverage for the remaining invalid-frame guard branches: frame without `shape`, and 1D `shape`

Validation:
- `python3 -m unittest tests.test_face_swapper_mask -v`

PR #1757, opened 2026-04-14
fix: guard against None best_face in default_target_face when no faces detected

### LINKED ISSUE #1755: fix: default_target_face crashes with TypeError when no faces detected in any frame
## Problem

In `modules/face_analyser.py`, the function `default_target_face()` has a bug that causes a `TypeError` (or `KeyError`) crash when a cluster/map entry contains no frames with any detected faces.

### Faulty Code

```python
def default_target_face():
    for map in modules.globals.source_target_map:
        best_face = None
        best_frame = None
        for frame in map['target_faces_in_frame']:
            if len(frame['faces']) > 0:
                best_face = frame['faces'][0]
                best_frame = frame
                break

        for frame in map['target_faces_in_frame']:
            for face in frame['faces']:
                if face['det_score'] > best_face['det_score']:  # <-- CRASH if best_face is None
                    best_face = face
                    best_frame = frame

        x_min, y_min, x_max, y_max = best_face['bbox']  # <-- also crashes if None
        ...
```

If a centroid cluster has no frames with detected faces (e.g. all faces dropped below the detection threshold, or the video segment has no detectable faces), `best_face` remains `None` after the first loop. The second loop then immediately raises:

```
TypeError: 'NoneType' object is not subscriptable
```

## Steps to Reproduce

1. Process a video where one face cluster contains only low-confidence/undetected faces.
2. Call `get_unique_faces_from_target_video()` which calls `default_target_face()`.
3. Observe crash:  `TypeError: 'NoneType' object is not subscriptable`

## Expected Behavior

The function should gracefully skip map entries where no face was detected, rather than crashing.

## Fix

Add a guard after the first loop:

```python
def default_target_face():
    for map in modules.globals.source_target_map:
        best_face = None
        best_frame = None
        for frame in map['target_faces_in_frame']:
            if len(frame['faces']) > 0:
                best_face = frame['faces'][0]
                best_frame = frame
                break

        if best_face is None:
            continue  # No faces detected in this cluster — skip

        for frame in map['target_faces_in_frame']:
            for face in frame['faces']:
                if face['det_score'] > best_face['det_score']:
                    best_face = face
                    best_frame = frame

        x_min, y_min, x_max, y_max = best_face['bbox']
        ...
```

kuishou68:
Closes #1755

## Problem
`default_target_face()` in `modules/face_analyser.py` crashes with `TypeError: 'NoneType' object is not subscriptable` when a face cluster has no frames with detected faces. `best_face` remains `None` after the first loop, then the second loop tries to access `best_face['det_score']`.

## Fix
Added an early `continue` guard after the first loop — if `best_face is None`, we skip this cluster entirely instead of crashing.

```python
if best_face is None:
    continue  # No faces detected in this cluster — skip
```

This is a one-line defensive guard that prevents the crash with no behavior change for the normal (faces detected) path.

## Summary by Sourcery

Bug Fixes:
- Avoid a TypeError in default_target_face when processing face clusters that contain no detected faces.

### DIFF
--- a/modules/face_analyser.py
+++ b/modules/face_analyser.py
@@ -179,6 +179,9 @@ def default_target_face():
                 best_frame = frame
                 break
 
+        if best_face is None:
+            continue  # No faces detected in this cluster — skip
+
         for frame in map['target_faces_in_frame']:
             for face in frame['faces']:
                 if face['det_score'] > best_face['det_score']:


noahximus:
Thanks for the focused fix. The guard safely handles empty face clusters without changing the normal detection path, and downstream mapping already tolerates entries without a selected target face. I reviewed the affected flow and this is good to merge.

PR #1771, opened 2026-04-16
Guard ONNX enhancer against malformed landmarks

duyua9:
## Summary
- convert the 106-point landmark fallback to a checked float array before indexing
- return an empty transform for missing, short, or malformed landmark data
- add lightweight tests for short, malformed, and unconvertible landmark inputs

## Tests
- python3 -m unittest tests/test_onnx_enhancer_affine.py
- python3 -m py_compile modules/processors/frame/_onnx_enhancer.py tests/test_onnx_enhancer_affine.py
- git diff --check

## Summary by Sourcery

Guard ONNX face affine computation against malformed 106-point landmark inputs and add tests to ensure invalid data returns empty transforms instead of reaching downstream processing.

Bug Fixes:
- Validate and safely convert 106-point landmark data to a float array, returning empty transforms for missing, short, or malformed inputs instead of attempting to index or process them.

Tests:
- Add unit tests covering short, malformed, and unconvertible 106-point landmark inputs to confirm they yield empty transforms and do not invoke OpenCV affine estimation.

### DIFF
--- a/modules/processors/frame/_onnx_enhancer.py
+++ b/modules/processors/frame/_onnx_enhancer.py
@@ -168,7 +168,14 @@ def _get_face_affine(face: Any, input_size: int):
     if hasattr(face, "kps") and face.kps is not None:
         landmarks = face.kps.astype(np.float32)
     elif hasattr(face, "landmark_2d_106") and face.landmark_2d_106 is not None:
-        lm106 = face.landmark_2d_106
+        try:
+            lm106 = np.asarray(face.landmark_2d_106, dtype=np.float32)
+        except (TypeError, ValueError):
+            return None, None
+
+        if lm106.ndim != 2 or lm106.shape[0] <= 88 or lm106.shape[1] < 2:
+            return None, None
+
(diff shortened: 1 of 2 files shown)


noahximus:
Can we implement whitelisting instead of blacklisting models?

duyua9:
Thanks for the guidance. I updated the fallback to use an allow-list for the exact InsightFace 106-point landmark layout instead of accepting arbitrary longer landmark arrays.

Changes made:
- Require `landmark_2d_106` to have exactly 106 rows and at least 2 columns before using the fallback.
- Centralized the accepted GPEN alignment indexes in `LANDMARK_2D_106_AFFINE_INDICES`.
- Added a regression test that rejects a non-106-point layout (107 rows) before affine estimation.

Validated with:
- `python3 -m unittest tests/test_onnx_enhancer_affine.py`
- `python3 -m py_compile modules/processors/frame/_onnx_enhancer.py tests/test_onnx_enhancer_affine.py`
- `git diff --check`

PR #1772, opened 2026-04-16
Guard lower mouth mask against invalid frames

duyua9:
## Summary
- validate frame input before reading frame.shape in create_lower_mouth_mask
- return the existing empty mask/default mouth metadata for None, missing-shape, and 1D frames
- extend the lightweight face swapper mask tests, including the import stub needed for insightface.utils.face_align

## Tests
- python3 -m unittest tests/test_face_swapper_mask.py
- python3 -m py_compile modules/processors/frame/face_swapper.py tests/test_face_swapper_mask.py
- git diff --check

## Summary by Sourcery

Guard lower mouth mask creation against invalid frame inputs and extend tests accordingly.

Bug Fixes:
- Prevent create_lower_mouth_mask from accessing frame.shape on None, non-array, or 1D frame inputs by returning the default empty mask and mouth metadata instead.

Tests:
- Extend lightweight face swapper mask tests to cover lower mouth mask behavior for None, missing-shape, and 1D frame inputs.
- Stub insightface.utils.face_align in tests to satisfy new or existing imports.

### DIFF
--- a/modules/processors/frame/face_swapper.py
+++ b/modules/processors/frame/face_swapper.py
@@ -1,4 +1,4 @@
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Tuple
 import cv2
 import insightface
 from insightface.utils import face_align
@@ -38,6 +38,8 @@ DETECTION_INTERVAL = 0.033  # ~30 FPS detection rate for live mode
 FRAME_SKIP_COUNTER = 0
 ADAPTIVE_QUALITY = True
 # --- END: Mac M1-M5 Optimizations ---
+
+EMPTY_MOUTH_BOX = (0, 0, 0, 0)
 
 abs_dir = os.path.dirname(os.path.abspath(__file__))
 models_dir = os.path.join(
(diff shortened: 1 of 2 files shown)


KRSHH:
Hi @duyua9 Thanks for the efforts, can you combine your Pull requests into few based on category and then share changes and benchmarks in each.

duyua9:
Thanks — that's a fair point.

You're right that these should be grouped more clearly by category instead of staying as many tiny PRs.

For the currently open fixes I would group them roughly like this:

1. **Invalid-input / guard fixes**
   - #1769
   - #1771
   - #1772

2. **Camera / media robustness fixes**
   - #1768
   - #1770
   - #1778

For these PRs, the relevant “benchmarks” are mostly **validation / behavior checks** rather than performance benchmarks, since they are correctness and robustness fixes.

For example, the validation I’ve been using is:
- reproduce the failing/unsafe input path before the patch
- verify the function now returns a stable fallback instead of crashing
- add a focused regression test for that exact edge case
- run the targeted test file locally

For `#1772` specifically:
- it guards `create_lower_mouth_mask` against `None`, missing-shape, and 1D frame inputs
- expected result after the fix is the existing empty-mask/default metadata path instead of an exception
- validation used:
  - `python3 -m unittest tests/test_face_swapper_mask.py`
  - `python3 -m py_compile modules/processors/frame/face_swapper.py tests/test_face_swapper_mask.py`
  - `git diff --check`

For `#1778` specifically:
- it makes `detect_fps()` return the existing `30.0` fallback when `ffprobe` fails or returns malformed FPS metadata
- validation used:
  - success case with valid `r_frame_rate`
  - subprocess failure
  - malformed output
  - zero denominator
  - `python3 -m unittest tests.test_utilities`

Going forward I can keep the fixes batched more cleanly by category and include the validation results explicitly in each PR. If you prefer, I can also consolidate the remaining open fixes further and close superseded ones.

duyua9:
Thanks again for the guidance.

I have rebased/merged this PR onto the latest `main` so it is no longer dirty. Current head is `526b7db` and GitHub now reports it as mergeable.

This PR remains in the **invalid-input / guard fixes** category:

- It only touches `create_lower_mouth_mask` and its focused regression coverage.
- It returns the existing empty-mask/default metadata path for invalid frame inputs instead of raising from `frame.shape` access.
- I kept the new lightweight regression test file even though latest `main` removed the previous test file, because it is the only focused validation for this guard behavior.

Validation run after resolving the latest `main` conflict:

- `python3 -m unittest tests/test_face_swapper_mask.py`
- `python3 -m py_compile modules/processors/frame/face_swapper.py tests/test_face_swapper_mask.py`
- `git diff --check`
- strict conflict-marker scan

No performance benchmark is expected here because this is an early invalid-input return path; the relevant check is that invalid inputs now return the documented empty defaults without entering the normal mask computation path.

PR #1776, opened 2026-04-22
Paste-back: O(crop_area) compositing + uint8 cv2 SIMD blend

maxwbuckley:
## Summary

Paste-back was the dominant face-size-dependent cost in the live pipeline: at large face bboxes the erode+blur in output coordinates ran kernel sizes proportional to the face, making the per-frame cost ~O(area · k²). This PR rewrites paste-back in both the face swapper and the GFPGAN enhancer so the per-frame cost is linear in the face crop area, with the feather baked into a cached alpha template at aligned-face resolution.

On an M-series Mac with CoreML, live preview FPS at a half-screen face went from ~20 to a steady 30 (camera-bound).

## Changes

- **`modules/processors/frame/face_swapper.py` — `_fast_paste_back`**
  - Precomputes a feathered alpha template once at aligned-face resolution (128×128 for inswapper); per frame it just warps the *soft* mask into the crop — no output-space erode or blur. Affine transform preserves the relative feather width, so the visual feather is unchanged.
  - Collapses the CPU fallback onto the same shape as the GPU path. Previously it did a full-frame `warpAffine` twice per call regardless of how much of the frame the face actually covered.
  - Swaps the CPU blend to a fused uint8 path (`cv2.merge` + `cv2.multiply` + `cv2.add`), dropping the numpy float32 round-trip. Micro-bench on a 1000×1000 crop: 9.4 ms → 1.2 ms (~8× faster, max pixel diff 2/255).

- **`modules/processors/frame/face_enhancer.py` — `_paste_back`**
  - Same cv2 SIMD blend pattern and uint8 alpha template. The enhancer already used a precomputed template, but in float32.

- **`modules/ui.py` — preview display path**
  - Processing thread now queues the frame as BGR; display thread resizes to the preview window first, then runs `cvtColor` on the ~5× smaller buffer.

## CPU compatibility (non-SIMD hardware)

The ~8× blend speedup leans on cv2's NEON/SSE uint8 path. On a hypothetical CPU without SIMD:

- **Correctness is unaffected.** `cv2.merge` / `cv2.multiply` / `cv2.add` all have scalar fallbacks and produce identical results. The max 2/255 delta vs. the old float32 path is from integer rounding in `cv2.multiply`, not from SIMD.
- **No regression.** The uint8 path touches ~¼ the memory of the old float32 path, so memory-bandwidth wins survive even with scalar fallback; the speedup just narrows from ~8× to ~2–3×.
- **Realistic hardware is fine.** Every x86 CPU since the Pentium 4 has SSE2, every ARMv8 (all Apple Silicon) has NEON by spec, and Rosetta translates SSE2 to NEON transparently. Anything old enough to lack SIMD can't run GFPGAN/inswapper in real time anyway.

No guard branch or feature flag is needed.

## Test plan

- [x] Live webcam preview on Apple Silicon (CoreML): verify steady FPS with face close to / far from the camera
- [ ] Live webcam preview on Windows with CUDA: verify no regression on the GPU blend path
- [ ] Live webcam preview with face enhancer enabled: verify no seam artifacts at the face edge
- [ ] Offline video processing: verify output matches pre-change (feather radius is preserved; pixel diff ≤ 2/255 from the blend precision change)

### DIFF
--- a/modules/processors/frame/face_enhancer.py
+++ b/modules/processors/frame/face_enhancer.py
@@ -178,17 +178,17 @@ def _paste_back(
     h, w = frame.shape[:2]
     inv_matrix = cv2.invertAffineTransform(affine_matrix)
 
-    # Build or reuse cached feathered mask
+    # Build or reuse cached feathered mask (uint8 — blended via cv2 SIMD ops)
     if _enhancer_cache['mask_size'] != output_size:
-        face_mask = np.ones((output_size, output_size), dtype=np.float32)
+        face_mask_f = np.ones((output_size, output_size), dtype=np.float32)
         border = max(1, int(output_size * 0.05))
         ramp_up = np.linspace(0.0, 1.0, border, dtype=np.float32)
         ramp_down = np.linspace(1.0, 0.0, border, dtype=np.float32)
-        face_mask[:border, :] *= ramp_up[:, None]
-        face_mask[-border:, :] *= ramp_down[:, None]
-        face_mask[:, :border] *= ramp_up[None, :]
-        face_mask[:, -border:] *= ramp_down[None, :]
-        _enhancer_cache['mask'] = face_mask
+        face_mask_f[:border, :] *= ramp_up[:, None]
+        face_mask_f[-border:, :] *= ramp_down[:, None]
+        face_mask_f[:, :border] *= ramp_up[None, :]
+        face_mask_f[:, -border:] *= ramp_down[None, :]
+        _enhancer_cache['mask'] = (face_mask_f * 255.0).astype(np.uint8)
         _enhancer_cache['mask_size'] = output_size
 
     # Compute tight bbox from affine corners (avoids full-frame warpAffine scan)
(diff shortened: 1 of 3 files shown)


maxwbuckley:
Thanks for the review. Responses:

**1. Non-square aligned face (`_get_soft_alpha` keyed by `face_h` only).** Valid defensive ask. Added an explicit `assert face_h == face_w` at the call site in `f95a0bb` — the inswapper model is fixed at 128×128 aligned-face space, but if that ever changes the mis-warp would be silent. Fail-loud is better.

**2. Thread safety of `_paste_cache` / `_enhancer_cache`.** In this codebase paste-back is only called from the single processing thread (`_processing_thread_func` in `modules/ui.py`). Even in `many_faces` mode, `swap_face` is invoked in a sequential loop per frame, not across threads. The caches are also effectively write-once: the `if size != cached_size` branch fires on the first call at a new size and never again (128 for inswapper, 512 for GFPGAN). So a lock would guard a one-shot init against a contention path that doesn't exist. I'd rather not add code for a non-race.

**3. BGR vs RGB queue semantics.** Checked: `processed_queue` is local to `create_webcam_preview` in `modules/ui.py`, with one producer (`_processing_thread_func`) and one consumer (`_display_next_frame`). No other code paths reference it. The BGR-instead-of-RGB switch is fully contained.

hacksider:
@maxwbuckley thanks again bro! Great performance gain! You're awesome!

PR #1778, opened 2026-04-22
fix: make detect_fps fall back on ffprobe errors

duyua9:
## Summary
- make `detect_fps()` fall back to `30.0` when `ffprobe` fails or returns malformed frame-rate output
- use explicit UTF-8 decoding for the `ffprobe` result
- add focused tests for valid output, command failure, malformed output, and zero denominator

## Why
`detect_fps()` currently calls `ffprobe` outside the protected parsing block, so subprocess failures can still bubble up and break video processing even though the function already has a `30.0` fallback.

This keeps the existing fallback behavior but applies it to command failures and malformed metadata as well.

## Testing
- `python3 -m unittest tests.test_utilities`

## Summary by Sourcery

Harden video frame rate detection to handle ffprobe failures and malformed metadata by falling back consistently to a default FPS value.

Bug Fixes:
- Ensure detect_fps falls back to a default FPS when ffprobe fails, returns malformed output, or produces an invalid fraction such as a zero denominator.

Enhancements:
- Decode ffprobe output explicitly as UTF-8 before parsing the frame rate.

Tests:
- Add unit tests covering normal fraction parsing, ffprobe command failures, malformed ffprobe output, and zero-denominator frame rate values.

### DIFF
--- a/modules/utilities.py
+++ b/modules/utilities.py
@@ -53,7 +53,16 @@ def detect_fps(target_path: str) -> float:
         target_path,
     ]
     try:
-        output = subprocess.check_output(command, encoding="utf-8").strip().split("/")
+        # Keep shell=False by passing a sequence; target_path is an ffprobe argument,
+        # not command text, so paths containing shell metacharacters are not executed.
+        completed = subprocess.run(
+            command,
+            check=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.DEVNULL,
+            text=True,
+        )
+        output = completed.stdout.strip().split("/")
         numerator, denominator = map(int, output)
         return numerator / denominator
     except (
--- a/modules/utilities.py
+++ b/modules/utilities.py
@@ -40,23 +40,22 @@ def run_ffmpeg(args: List[str]) -> bool:
 
 
 def detect_fps(target_path: str) -> float:
-    command = [
-        "ffprobe",
-        "-v",
-        "error",
-        "-select_streams",
-        "v:0",
-        "-show_entries",
-        "stream=r_frame_rate",
-        "-of",
-        "default=noprint_wrappers=1:nokey=1",
-        target_path,
-    ]
     try:
-        # Keep shell=False by passing a sequence; target_path is an ffprobe argument,
-        # not command text, so paths containing shell metacharacters are not executed.
+        # Keep shell=False by passing a literal argument vector; target_path is an
+        # ffprobe argument, not command text, so shell metacharacters are inert.
         completed = subprocess.run(
-            command,
+            [
+                "ffprobe",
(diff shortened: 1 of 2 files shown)


duyua9:
Adding the concrete validation results here as well in case that's helpful for review.

This PR is intended as a robustness fix, so the useful signal is behavior validation rather than a performance benchmark.

Validated cases:
- valid `ffprobe` output like `30000/1001` still returns the parsed FPS
- `ffprobe` command failure now returns the existing `30.0` fallback instead of bubbling up an exception
- malformed output now returns `30.0`
- zero-denominator output now returns `30.0`

Targeted test run:
- `python3 -m unittest tests.test_utilities`

So the expected before/after behavior is:
- **before**: subprocess/parsing failures in `detect_fps()` could still break the video-processing path
- **after**: the function keeps the existing fallback contract and degrades to `30.0` for invalid or unavailable FPS metadata

duyua9:
Follow-up pushed in `a869d5b` for the Sourcery security warning.\n\nI replaced `subprocess.check_output` with `subprocess.run(..., shell=False, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL, text=True, check=True)` and kept the ffprobe invocation as an argument list. The code now documents that `target_path` is passed as an ffprobe argument rather than shell command text, so shell metacharacters in paths are not executed. The regression tests were updated to mock `subprocess.run`.\n\nValidation run locally:\n- `python3 -m pytest -q tests/test_utilities.py`\n- `python3 -m py_compile modules/utilities.py tests/test_utilities.py`\n- `git diff --check`

duyua9:
Second follow-up pushed in `2ed5106` after Sourcery re-ran and still flagged `subprocess.run(command, ...)` because the first argument was stored in a variable.\n\nI inlined the ffprobe argument vector directly in `subprocess.run([...], shell=False, ...)` so the executable/options are visible as a literal argument list at the call site while `target_path` remains a single inert argument.\n\nValidation re-run locally:\n- `python3 -m pytest -q tests/test_utilities.py`\n- `python3 -m py_compile modules/utilities.py tests/test_utilities.py`\n- `git diff --check`

duyua9:
Second follow-up pushed in `2ed5106` after Sourcery re-ran and still flagged `subprocess.run(command, ...)` because the first argument was stored in a variable.\n\nI inlined the ffprobe argument vector directly in `subprocess.run([...], shell=False, ...)` so the executable/options are visible as a literal argument list at the call site while `target_path` remains a single inert argument.\n\nValidation re-run locally:\n- `python3 -m pytest -q tests/test_utilities.py`\n- `python3 -m py_compile modules/utilities.py tests/test_utilities.py`\n- `git diff --check`

PR #1783, opened 2026-04-23
fix: no cam detected on Arch Linux, use string path for camera capture instead

Vito-M:
On the latest version of Arch Linux, the camera was no longer being detected at startup, causing the application to fail to list any available capture devices.

The fix passes the device path as a string (e.g. `/dev/video0`) directly, which resolves the issue on Arch Linux while preserving the existing behavior on other platforms (Windows/macOS).

## Summary by Sourcery

Update Linux camera handling to open devices via their /dev/videoN path instead of numeric indices and keep existing behavior for other platforms.

Bug Fixes:
- Ensure cameras are correctly detected and opened on recent Arch Linux systems by passing explicit /dev/videoN paths to OpenCV.

Enhancements:
- Refine platform-specific video capture setup to distinguish Linux device-path usage from macOS auto-detection behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ tf_env/
 *.mp4
 *.mkv
 
+.python-version
 .tmp/
 temp/
 .venv/
--- a/modules/ui.py
+++ b/modules/ui.py
@@ -1065,7 +1065,7 @@ def get_available_cameras():
         else:
             # Linux camera detection - test first 10 indices
             for i in range(10):
-                cap = cv2.VideoCapture(i)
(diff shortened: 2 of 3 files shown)


noahximus:
This looks like a solid, narrowly-scoped fix — thanks. It's conflicting with main now though, most likely from other recent changes to video_capture.py's camera-backend selection (there's a parallel macOS AVFoundation fix in flight in #1884 touching the same function). Could you rebase and make sure the Linux and macOS branches stay separate? Once that's clean this is good to go.

PR #1798, opened 2026-05-01
Download GFPGAN enhancer model when missing

zuyua9:
## What

Make the default GFPGAN face enhancer download its required ONNX model when it is missing.

The GPEN enhancers and the face swapper already use `conditional_download()` for their model files, but `face_enhancer.py` only reported that `models/gfpgan-1024.onnx` had to be placed manually. This makes the default enhancer fail during pre-check even though the project already has the model downloader path used by adjacent processors.

This PR:

- defines the GFPGAN model filename and Hugging Face model URL in `face_enhancer.py`
- calls `conditional_download()` from `pre_check()` when the model is missing
- retries the same download during lazy ONNX session initialization before raising `FileNotFoundError`
- keeps the existing local-file preference: if `models/gfpgan-1024.onnx` already exists, no download is attempted

## Validation

- `python3 -m py_compile modules/processors/frame/face_enhancer.py`
- `git diff --check`
- targeted stub import test: monkey-patched the heavy runtime imports and verified `pre_check()` calls `conditional_download(models_dir, [MODEL_URL])` for a missing model and succeeds after the file appears
- verified the model URL resolves with an HTTP HEAD request

## Summary by Sourcery

Ensure the default GFPGAN face enhancer automatically retrieves its required ONNX model instead of requiring manual placement.

New Features:
- Add automatic download of the GFPGAN ONNX model when it is missing for the face enhancer processor.

Enhancements:
- Centralize the GFPGAN model filename and Hugging Face URL as constants shared by pre-check and runtime initialization.
- Improve error messaging when the GFPGAN model is not found, clarifying download or manual placement options.

### DIFF
--- a/modules/processors/frame/face_enhancer.py
+++ b/modules/processors/frame/face_enhancer.py
@@ -1,6 +1,6 @@
 # Uses ONNX Runtime for GFPGAN face enhancement (no torch/gfpgan dependency)
 
-from typing import Any, List
+from typing import Any, List, Optional
 import cv2
 import threading
 import numpy as np
@@ -48,19 +48,26 @@ FFHQ_TEMPLATE_512 = np.array(
 
 
 def pre_check() -> bool:
+    model_path = ensure_model_available(report_status=True)
+    return model_path is not None
+
+
+def ensure_model_available(report_status: bool = False) -> Optional[str]:
     model_path = os.path.join(models_dir, MODEL_FILE)
     if not os.path.exists(model_path):
-        update_status(f"Downloading {MODEL_FILE}...", NAME)
+        if report_status:
+            update_status(f"Downloading {MODEL_FILE}...", NAME)
         conditional_download(models_dir, [MODEL_URL])
 
     if not os.path.exists(model_path):
-        update_status(
-            f"GFPGAN ONNX model not found at {model_path}. "
-            f"Download {MODEL_FILE} or place it in the models folder.",
-            NAME,
-        )
-        return False
-    return True
+        if report_status:
+            update_status(
+                f"Automatic download failed for {MODEL_FILE} from {MODEL_URL}. "
+                f"Place the model at {model_path} and try again.",
+                NAME,
+            )
+        return None
+    return model_path
 
 
 def pre_start() -> bool:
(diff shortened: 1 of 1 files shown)


zuyua9:
Thanks for the review. I pushed a follow-up commit that addresses both Sourcery suggestions:

- extracted the repeated GFPGAN model path/download/existence logic into `ensure_model_available()` so `pre_check()` and `get_face_enhancer()` share the same behavior;
- updated the failure messages to say that the automatic download failed and include the URL / expected local path, instead of only telling the user to download/place the model manually.

Validation:

- `python3 -m py_compile modules/processors/frame/face_enhancer.py`
- `git diff --check`

modules/processors/frame/face_enhancer.py
noahximus, line 33: Can you do something like this instead for better readability 
'''
base_dir = os.path.abspath(os.path.join(abs_dir, "../../.."))
models_path = os.path.join(base_dir, "models")
'''

modules/processors/frame/face_enhancer.py
```diff
@@ -43,15 +48,26 @@
...
+
+    if not os.path.exists(model_path):
+        if report_status:
+            update_status(
+                f"Automatic download failed for {MODEL_FILE} from {MODEL_URL}. "
+                f"Place the model at {model_path} and try again.",
+                NAME,
+            )
+        return None
+    return model_path
```
noahximus, line 70: Can you do it this way instead 
'''
def ensure_model_available(report_status: bool = False) -> Optional[str]:
    model_path = os.path.join(models_dir, MODEL_FILE)

    if os.path.isfile(model_path):
        return model_path

    if report_status:
        update_status(f"Downloading {MODEL_FILE}...", NAME)

    try:
        os.makedirs(models_dir, exist_ok=True)
        conditional_download(models_dir, [MODEL_URL])
    except Exception as exc:
        if report_status:
            update_status(
                f"Download error for {MODEL_FILE}: {exc}. "
                f"Place the model at {model_path} and try again.",
                NAME,
            )
        return None

    if not os.path.isfile(model_path):
        if report_status:
            update_status(
                f"Automatic download failed for {MODEL_FILE}. "
                f"Place the model at {model_path} and try again.",
                NAME,
            )
        return None

    return model_path
'''

amainooti:
Why don't you also include this in the instruction and provide a link for the gfpgan-1024.onnx 
https://github.com/Glat0s/GFPGAN-1024-onnx/releases/download/v0.0.1/gfpgan-1024.onnx

Artcmx69 (approved):

PR #1803, opened 2026-05-07
fix(face): reuse pre-detected face list

zuyua9:
## Summary
- allow `get_one_face()` to reuse an optional list of already-detected faces
- preserve the existing frame-analysis path when no face list is supplied
- add focused tests for supplied faces, empty supplied faces, and the existing detection path

## Why
`modules/processors/frame/face_swapper.py` already calls `get_one_face(processed_frame, detected_faces)` in the fallback path after computing `detected_faces`. On current `main`, `get_one_face()` only accepts one positional argument, so that path raises `TypeError` before it can reuse the detected faces.

This keeps the existing behavior for callers that only pass a frame, while avoiding an extra heavy analysis pass when the caller already has the face list.

## Validation
- `python3 -m unittest tests/test_face_analyser_get_one_face.py -v`
- `python3 -m py_compile modules/face_analyser.py tests/test_face_analyser_get_one_face.py`
- `git diff --cached --check` before commit

## Summary by Sourcery

Allow reusing a pre-computed face list in face selection while preserving existing behavior when no list is provided

New Features:
- Support passing an optional list of detected faces into get_one_face() to select a face without re-analysing the frame

Bug Fixes:
- Prevent TypeError in face_swapper fallback path by aligning get_one_face() signature with its usage when a faces list is supplied

Tests:
- Add unit tests covering use of supplied detected faces, empty supplied face lists, and the legacy detection path when no faces are provided

### DIFF
--- a/modules/face_analyser.py
+++ b/modules/face_analyser.py
@@ -138,15 +138,16 @@ def _analyse_faces(frame: Frame) -> list:
     return faces
 
 
-def get_one_face(frame: Frame) -> Any:
-    if _is_dml():
-        with modules.globals.dml_lock:
+def get_one_face(frame: Frame, faces: Any = None) -> Any:
+    if faces is None:
+        if _is_dml():
+            with modules.globals.dml_lock:
+                faces = _analyse_faces(frame)
+        else:
             faces = _analyse_faces(frame)
-    else:
-        faces = _analyse_faces(frame)
     try:
         return min(faces, key=lambda x: x.bbox[0])
-    except ValueError:
+    except (TypeError, ValueError):
         return None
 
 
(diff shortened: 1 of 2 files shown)


zuyua9:
Addressed Sourcery’s review feedback in d1376b07d187b4bddaf3600cc658f778da4a3295: `get_one_face()` now only catches `ValueError` for the empty-face-list case, so invalid/non-iterable face inputs still surface as errors instead of being hidden. Re-ran `python3 -m unittest tests/test_face_analyser_get_one_face.py -v`, `python3 -m py_compile modules/face_analyser.py tests/test_face_analyser_get_one_face.py`, and `git diff --check`.

PR #1812, opened 2026-05-09
fix(cli): allow disabling audio restore

euyua9:
## Summary
- add a `--no-keep-audio` CLI flag so headless runs can skip audio restoration
- keep the existing `--keep-audio` default behavior unchanged
- document the inverse flag in the README options list
- add a lightweight regression test that avoids importing the heavy runtime dependencies

## Tests
- `python3 -m pytest tests/test_cli_keep_audio_flag.py -q`
- `python3 -m py_compile modules/core.py`
- `git diff --check origin/main...HEAD`

## Summary by Sourcery

Add a CLI option to disable audio restoration while preserving the existing default behavior and document and test the new flag.

New Features:
- Introduce a --no-keep-audio CLI flag to allow disabling original audio restoration in headless runs.

Documentation:
- Document the --no-keep-audio option in the README CLI options list.

Tests:
- Add regression tests to verify the presence and behavior of the keep-audio and no-keep-audio CLI flags.


euyua9:
Thanks for the review. I pushed a follow-up that addresses both points:

- The AST test now resolves `modules/core.py` relative to `__file__`, so it no longer depends on pytest being invoked from the repository root.
- The audio flag default is now set on the mutually exclusive group with `audio_group.set_defaults(keep_audio=True)`, while `--keep-audio` and `--no-keep-audio` only set the explicit overrides.

Validation:
- `python3 -m pytest tests/test_cli_keep_audio_flag.py -q` — 2 passed
- `(cd /tmp && python3 -m pytest /Users/ming/code/pr-worktrees/deep-live-cam-no-keep-audio-euyua9/tests/test_cli_keep_audio_flag.py -q)` — 2 passed
- `python3 -m py_compile modules/core.py`
- `git diff --check origin/main...HEAD`

PR #1822, opened 2026-05-14
fix: validate in-memory pipe completion

euyua9:
Summary:
- Treat FFmpeg decoder failures as in-memory pipeline failures instead of accepting a partial output file
- Reject zero-frame and truncated in-memory runs so the caller can fall back to the disk pipeline
- Add focused tests for decoder failure, truncated output, and a complete successful pipe run

Validation:
- `python3 -m pytest -q tests/test_face_analyser_get_one_face.py tests/test_pipe_pipeline_integrity.py --tb=short`
- `git diff --check`

## Summary by Sourcery

Validate in-memory FFmpeg pipe runs by treating decoder errors, zero-frame output, and truncated frame processing as pipeline failures instead of successes.

Bug Fixes:
- Treat non-zero FFmpeg decoder return codes as in-memory pipeline failures so partial outputs are not accepted.
- Reject in-memory runs that process zero frames or fewer frames than expected, allowing callers to fall back to the disk-based pipeline.

Tests:
- Add focused unit tests covering decoder failure, truncated output, and a successful complete in-memory pipeline run.

### DIFF
--- a/modules/processors/frame/core.py
+++ b/modules/processors/frame/core.py
@@ -383,13 +383,29 @@ def _run_pipe_pipeline(
         writer.wait()
         reader.wait()
 
+        if reader.returncode != 0:
+            stderr_out = reader.stderr.read().decode(errors='ignore').strip()
+            if stderr_out:
+                print(f"[DLC.CORE] FFmpeg decoder error: {stderr_out}")
+            return False
+
         if writer.returncode != 0:
             stderr_out = writer.stderr.read().decode(errors='ignore').strip()
             if stderr_out:
                 print(f"[DLC.CORE] FFmpeg encoder error: {stderr_out}")
             return False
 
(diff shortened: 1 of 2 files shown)


kier007:
Thanks for the PR. I’m going to reject this for now.

The direction makes sense, but I don’t think this is ready to merge yet because the failure coverage is incomplete. This changes the success criteria for the in-memory FFmpeg pipeline, but the tests only cover decoder failure, truncation, and a successful run. The encoder failure path should also be covered so we don’t regress into accepting partial or invalid output.

I’m also not fully comfortable calling the new test file “pipeline integrity” coverage because the tests pre-create the output file rather than verifying that FFmpeg actually produced a valid output. That makes the tests useful for return-code/frame-count behavior, but weaker than the PR description suggests.

Please add coverage for writer/encoder failure and make sure the validation command passes in CI before this is reconsidered.

PR #1790, opened 2026-04-27
Quality: Invalid frame index calculation can seek to -1 or out-of-range frames

Nam0101:
### Problem
`get_video_frame()` computes the seek position as `min(frame_total, frame_number - 1)`. With the default `frame_number=0`, this becomes `-1` (invalid frame index). For large requested frame numbers, it can also become `frame_total` (also invalid, since last valid frame is `frame_total - 1`). This causes intermittent failed reads (`has_frame=False`) and breaks callers that expect a valid first frame by default.

**Severity**: `high`
**File**: `modules/capturer.py`

### Solution
Clamp to a valid range explicitly and avoid negative/out-of-bounds indices:
`frame_total = int(capture.get(cv2.CAP_PROP_FRAME_COUNT))`
`if frame_total <= 0:` `    capture.release()` `    return None`
`# keep 1-based API: frame_number=1 -> first frame` `target_index = 0 if frame_number <= 1 else min(frame_total - 1, frame_number - 1)` `capture.set(cv2.CAP_PROP_POS_FRAMES, target_index)`

### Changes
- `modules/capturer.py` (modified)

### Testing
- [x] Existing tests pass
- [x] Manual review completed
- [x] No new warnings/errors introduced

---

---

<details>
<summary>🤖 About this PR</summary>

This pull request was generated by [ContribAI](https://github.com/tang-vu/ContribAI), an AI agent
that helps improve open source projects. The change was:

1. **Discovered** by automated code analysis
2. **Generated** by AI with context-aware code generation
3. **Self-reviewed** by AI quality checks

If you have questions or feedback about this PR, please comment below.
We appreciate your time reviewing this contribution!

</details>

Closes #1789

## Summary by Sourcery

Bug Fixes:
- Prevent get_video_frame from seeking to negative or out-of-range frame indices, ensuring the default call returns a valid first frame and large frame requests clamp to the last valid frame.

PR #1824, opened 2026-05-15
fix: extract frames for map faces fallback

cuyua9:
## Summary

- always create the temp frame directory and extract frames before the disk-based fallback runs
- preserves the in-memory pipeline for non-map-faces runs, but lets `--map-faces` use the disk path correctly
- adds a regression test for the `map_faces` fallback path so frame extraction happens before `process_video`

## Tests

- `python3 -m unittest tests/test_core_map_faces_fallback.py -v`
- `python3 -m py_compile modules/core.py tests/test_core_map_faces_fallback.py`
- `git diff --check`

## pr-value

- pre-submit-final-diff: `pr-value-r=86`, `pr-value-l=100`, gate passed

## Summary by Sourcery

Ensure video frames are always extracted to a temporary directory before disk-based processing runs, including when map-faces fallback is used.

Bug Fixes:
- Fix map-faces disk fallback so it uses the frame-extraction path instead of skipping frame creation.

Tests:
- Add a regression test verifying that the map-faces disk fallback extracts frames before invoking video processing.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -276,10 +276,9 @@ def start() -> None:
             update_status('Falling back to disk-based processing...')
 
         extraction_start = time.time()
-        if not modules.globals.map_faces:
-            create_temp(modules.globals.target_path)
-            update_status('Extracting frames...')
-            extract_frames(modules.globals.target_path)
+        create_temp(modules.globals.target_path)
+        update_status('Extracting frames...')
+        extract_frames(modules.globals.target_path)
         extraction_time = time.time() - extraction_start
 
         temp_frame_paths = get_temp_frame_paths(modules.globals.target_path)
--- a/modules/core.py
+++ b/modules/core.py
@@ -270,10 +270,9 @@ def start() -> None:
             update_status('Falling back to disk-based processing...')
 
         extraction_start = time.time()
-        if not modules.globals.map_faces:
-            create_temp(modules.globals.target_path)
-            update_status('Extracting frames...')
-            extract_frames(modules.globals.target_path)
+        create_temp(modules.globals.target_path)
+        update_status('Extracting frames...')
+        extract_frames(modules.globals.target_path)
         extraction_time = time.time() - extraction_start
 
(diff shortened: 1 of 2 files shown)


noahximus:
Thanks, I tested the behavior and the fix is valid: with map_faces=True, current main skips frame extraction but still passes temp frame paths into process_video. This PR correctly creates temp resources and extracts frames before processing.

However, the patch no longer applies cleanly to current main, so I can’t merge it as-is. Could you please rebase/update the PR against current main? The intended production change still looks good.

cuyua9:
@noahximus thanks for the validation! I went to do the rebase you asked for and found the PR is already based on current `main` HEAD (`8234965 fix: clamp video frame seek index (#1790)`):

- PR base sha: `8234965ee8711394ccc47cadde8754b17cb34ee9`
- Current main HEAD: `8234965ee8711394ccc47cadde8754b17cb34ee9`
- `git rebase origin/main` ⇒ "Current branch is up to date", no SHAs rewritten
- `git merge-tree --write-tree --merge-base=$(git merge-base HEAD origin/main) origin/main HEAD` ⇒ clean tree, no conflicts
- GitHub PR API now reports `mergeable: true, rebaseable: true`

The "no longer applies cleanly" state you saw on 2026-07-23 was almost certainly a stale `mergeable_state` cache — the new main commit (`8234965`) had landed ~7h before your check, and GitHub's background merger can lag behind that for a while.

For completeness on the Sourcery suggestions from May: both were already applied in `368e840 test: isolate map faces fallback stubs` — `_patched_core_import_stubs` is a `@contextmanager` wrapping `patch.dict(sys.modules, …, clear=False)`, and the test now asserts the full pipeline ordering (`create_temp < extract_frames < process_video < create_video < move_temp`). Local run: `Ran 2 tests in 0.014s OK`.

Could you re-trigger the mergeability check (just re-opening the PR tab usually forces GitHub to recompute) and try merging? Happy to do an empty refresh push on the branch if that helps unstick the cache on your side.

noahximus:
@sourcery-ai review

PR #1825, opened 2026-05-15
fix: measure camera fps without dropping startup frames

duyua9:
## Summary

- Avoid consuming a warmup/sample burst in `VideoCapturer.start()` before live preview can read frames.
- Keep the immediate reported-FPS startup value, then refine `actual_fps` opportunistically from frames already returned by `read()`.
- Add regression tests proving startup does not call `read()` for FPS probing and that delivered frames still update `actual_fps`.

## Tests

- `python3 -m unittest tests/test_video_capture_fps_probe.py -v`
- `python3 -m py_compile modules/video_capture.py tests/test_video_capture_fps_probe.py`
- `git diff --check`

## Summary by Sourcery

Update video capture FPS handling to avoid consuming startup frames while still refining actual frame rate during normal reads.

New Features:
- Add opportunistic FPS measurement that refines actual_fps based on frames returned by read() rather than a dedicated sampling burst.

Bug Fixes:
- Prevent VideoCapturer.start() from consuming and discarding initial camera frames when probing FPS, preserving frames for live preview.

Tests:
- Add unit tests to verify start() no longer performs read()-based FPS probing and that actual_fps is updated from delivered frames.

### DIFF
--- a/modules/video_capture.py
+++ b/modules/video_capture.py
@@ -167,12 +167,14 @@ class VideoCapturer:
                 return
 
             elapsed = now - self._fps_sample_started_at
-            if elapsed > 0:
-                # N frames contain N-1 frame intervals from the first timestamp
-                # to the current one.
-                self.actual_fps = (self._fps_sample_count - 1) / elapsed
+            if elapsed <= 0:
+                return
+
+            # N frames contain N-1 frame intervals from the first timestamp
+            # to the current one.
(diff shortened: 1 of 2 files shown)


kier007:
not ready yet, `WebcamPreviewWindow` snapshots `actual_fps` before sampling and `_ProcessingWorker` keeps the stale value, so the measured FPS never changes polling or detection cadence. please wire the measured value through and make the capture tests stub `pygrabber` on Windows

PR #1830, opened 2026-05-18
fix: derive default threads from selected provider

euyua9:
## Summary
- compute the default execution provider once before registering CLI defaults
- derive the default execution thread count from that selected provider instead of the still-empty global provider list
- add regression coverage for CUDA and CPU default parsing paths

## Tests
- `python3 -m pytest tests/test_core_parse_args_defaults.py tests/test_face_analyser_get_one_face.py -q`

## Summary by Sourcery

Align default execution threads with the initially selected execution provider when parsing core CLI arguments.

Bug Fixes:
- Fix default execution thread selection to use the suggested default execution provider instead of an empty global provider list.

Tests:
- Add regression tests to verify default execution provider and thread count for CUDA and CPU configurations when parsing CLI arguments.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -56,9 +56,10 @@ def parse_args() -> None:
     program.add_argument('-l', '--lang', help='Ui language', default="en")
     program.add_argument('--live-mirror', help='The live camera display as you see it in the front-facing camera frame', dest='live_mirror', action='store_true', default=False)
     program.add_argument('--live-resizable', help='The live camera frame is resizable', dest='live_resizable', action='store_true', default=False)
+    default_execution_provider = suggest_default_execution_provider()
     program.add_argument('--max-memory', help='maximum amount of RAM in GB', dest='max_memory', type=int, default=suggest_max_memory())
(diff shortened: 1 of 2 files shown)


euyua9:
Added follow-up test coverage in 7738c5f for the provider-specific default execution-thread behavior. The test helper now restores temporary module state after each import, and DML/ROCM defaults are covered explicitly. Verified with: python -m pytest -q tests/test_core_parse_args_defaults.py (4 passed).

euyua9:
Rebuilt the branch on current main and force-updated it to 35c68f9 so the PR no longer carries the stale snapshot history. The runtime behavior is already present on current main; this commit adds focused coverage for CUDA, CPU, DML, and ROCM defaults while restoring isolated module state between tests. Verification: python -m pytest -q tests/test_core_parse_args_defaults.py (4 passed).

kier007:
not ready yet, the focused tests pass but `tests/test_core_parse_args_defaults.py` leaves `SIGINT` replaced after `parse_args()`, please isolate that side effect and update the title and body to match the test-only diff

PR #1832, opened 2026-05-19
feat: selectable GFPGAN model (1024 / 512) with hot-swap

5uck1ess:
## Summary

Lets the user pick between the two GFPGAN ONNX variants the project hosts via the existing Face Enhancer dropdown, and hot-swap between them without restarting. The single hardcoded `gfpgan-1024.onnx` load is replaced by a `globals.gfpgan_model_filename` lookup with `gfpgan-1024.onnx` as default (preserves current behavior).

## Why

The two GFPGAN ONNX models on the [hacksider/deep-live-cam HF repo](https://huggingface.co/hacksider/deep-live-cam/tree/main) have meaningfully different quality/speed trade-offs:

- **`gfpgan-1024.onnx`** (~365 MB) — 512 in / 1024 out (super-res head). Highest visual quality. Slower per call. Current default.
- **`GFPGANv1.4.onnx`** (~340 MB) — 512 in / 512 out. Standard GFPGAN 1.4 weights converted to ONNX. Lighter, faster.

Pre-PR there was no way to select the faster one without editing source. With both options exposed, users on lower-end GPUs (or USB-bandwidth-constrained webcams where the quality gain is invisible anyway) can pick the lighter model. Power users keep the 1024 head as default.

Both files are already hosted on the project's HF repo — no new external dependency, just an existing-but-unused asset becoming UI-selectable.

## What changed

- **`modules/globals.py`** — new `gfpgan_model_filename` global, defaults to `\"gfpgan-1024.onnx\"`. Comment notes both files are on the HF repo.
- **`modules/processors/frame/face_enhancer.py`**
  - `_gfpgan_model_path()` reads the global with a safe fallback.
  - `pre_check()` and `get_face_enhancer()` now use the helper instead of a hardcoded filename. Error message refers to the configured file.
  - New `reset_face_enhancer()` drops the cached ORT session and invalidates the feathered-mask cache so the next call reloads cleanly with the new model. Safe to call mid-session.
  - One-line log on session load (`Loading GFPGAN ONNX model from <path>`) so users can verify which variant is active — matches the existing GPEN-512 logging style.
- **`modules/ui.py`**
  - Enhancer dropdown options expanded:
    - Before: `[None, GFPGAN, GPEN-512, GPEN-256]`
    - After: `[None, GFPGAN-1024, GFPGAN-512, GPEN-512, GPEN-256]`
  - Initial value picks the variant matching the saved `gfpgan_model_filename`.
  - `_on_enhancer_change` maps both GFPGAN labels to the same `face_enhancer` processor key, sets the filename in globals, and calls `reset_face_enhancer()` only when the variant actually changed.

`enhance_face()` already auto-discovers the model's I/O shape from the ONNX session, so switching between 512-out and 1024-out is dimension-safe with no other code changes.

## Test plan

- [ ] Default state: launch with no prior settings → dropdown shows `GFPGAN-1024` when enhancer toggled on; behavior identical to pre-PR
- [ ] Place `GFPGANv1.4.onnx` from HF in `models/`, switch GFPGAN-1024 → GFPGAN-512 mid-session → console prints new model path, next swap uses 512-out
- [ ] Switch GFPGAN-512 → GFPGAN-1024 mid-session → reverse of above
- [ ] GPEN-512 / GPEN-256 paths unchanged from current behavior
- [ ] Missing `GFPGANv1.4.onnx` on disk → clear error referencing the missing filename (not the hardcoded 1024 message)

No new dependencies.

### DIFF
--- a/modules/globals.py
+++ b/modules/globals.py
@@ -70,6 +70,13 @@ enable_interpolation: bool = True # Toggle temporal smoothing
 interpolation_weight: float = 0  # Blend weight for current frame (0.0-1.0). Lower=smoother.
 # --- END: Added for Frame Interpolation ---
 
+# GFPGAN model filename. Currently shipped variants:
+#   gfpgan-1024.onnx  — 512 in, 1024 out (super-res head). Highest quality.
+#   GFPGANv1.4.onnx   — 512 in, 512 out. Lighter, faster.
+# Both files are hosted on the hacksider/deep-live-cam HF model repo.
+# Hot-swap supported via reset_face_enhancer() (called from UI on change).
+gfpgan_model_filename: str = "gfpgan-1024.onnx"
+
 # --- END OF FILE globals.py ---
 
 import threading
--- a/modules/processors/frame/face_enhancer.py
+++ b/modules/processors/frame/face_enhancer.py
@@ -55,9 +55,9 @@ def reset_face_enhancer() -> None:
     global FACE_ENHANCER
     with THREAD_LOCK:
         FACE_ENHANCER = None
-    # Also invalidate the feathered-mask cache — output size will likely change
-    _enhancer_cache['mask'] = None
-    _enhancer_cache['mask_size'] = 0
+        # Also invalidate the feathered-mask cache — output size will likely change
+        _enhancer_cache['mask'] = None
+        _enhancer_cache['mask_size'] = 0
 
 
 def pre_check() -> bool:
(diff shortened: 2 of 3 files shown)


kier007:
Good concept and pr desc, but I wouldn't approve it yet for 3 reasons:
- When selecting an absent model raises outside error handling and can terminate the webcam worker (`face_enhancer.py (line 96)`, `ui.py (line 1121)`)
- The variant isn't save, so `GFPGAN-512` reverts to 1024 after restart (`ui.py (line 299)`)
- `reset_face_enhancer()` clears a mask `_paste_back()` may be using it; the lock doesn't protect readers. It also leaves the live temporal cache stale (`face_enhancer.py (line 50)`)

PR #1833, opened 2026-05-19
feat: capture resolution + detection size dropdowns

5uck1ess:
## Summary

Adds two dropdowns to the Camera card on the main window:

- **Resolution** — requested webcam capture size (640×480 / 960×540 / 1280×720 / 1920×1080). Applied on next Live start.
- **Det size** — face detection input size (160 / 320 / 640). Triggers face analyser re-init so the change takes effect immediately.

Also exposes `--det-size` as a CLI flag for headless runs.

## Why

**Resolution:** Previously `VideoCapturer.start()` was hardcoded to `PREVIEW_DEFAULT_WIDTH/HEIGHT` which mixes \"the preview window's pixel size\" with \"the camera's capture resolution\" — they're separate concerns. With this PR they're decoupled and the capture side becomes user-configurable.

Why these choices specifically: 640×480 is native on virtually every USB 2.0 webcam and the safest 30fps mode (USB 2.0 isochronous bandwidth caps ~30fps on common YUYV at this size). Higher tiers are useful for users with capture cards or USB 3.0 webcams; the tooltip explains that 720p+ typically drops to ~10fps over USB 2.0 due to bandwidth, so users know what they're trading.

**Det size:** `insightface` defaults to 640×640 detection which is plenty accurate but expensive per frame. At 320 or 160 the detector runs much faster — a meaningful speedup when the face is large in frame (live webcam usage), with negligible accuracy loss for that distance regime. Currently the only way to change `det_size` is by editing the source.

## What changed

- **`modules/globals.py`**
  - New `det_size: int = 640` and `capture_resolution: tuple = (640, 480)` globals with documentation.
- **`modules/face_analyser.py`**
  - The hardcoded `DET_SIZE = (640, 640)` constant is replaced by `_current_det_size()` that reads `modules.globals.det_size` per call.
  - New `reset_face_analyser()` clears the cached `FACE_ANALYSER` so the next call re-prepares insightface with the new size.
- **`modules/core.py`** — new `--det-size` argparse flag with `choices=[160, 320, 640]` for headless runs.
- **`modules/ui.py`**
  - Camera card now uses a `QGridLayout` (was a single row) with the two new dropdowns on rows 1 and 2 below the camera selector + Live button.
  - `_on_resolution_change` writes the selected `(w, h)` tuple to `modules.globals.capture_resolution`.
  - `_on_det_size_change` writes the selected `int` and calls `reset_face_analyser()` so the new size kicks in on the next analyser call.
  - `WebcamPreviewWindow` reads `modules.globals.capture_resolution` instead of `PREVIEW_DEFAULT_WIDTH/HEIGHT` (the constants still drive preview-window sizing, separate concern).

## Test plan

- [ ] Defaults: launch, Camera card shows Resolution=`640x480` and Det size=`640 x 640`. Live mode behaves as before
- [ ] Switch Resolution to `1280 x 720`, start Live → console shows `[VideoCapturer] 1280x720` or the camera's nearest supported size
- [ ] Switch Det size from 640 → 320 while no Live is running → next face-analysis call (e.g. starting Live with map_faces) uses the new size
- [ ] Switch Det size mid-Live → analyser re-inits on the next frame; no crash
- [ ] `python run.py --det-size 320` headless honors the flag

No new dependencies.

## Summary by Sourcery

Add configurable camera capture resolution and face detection size for both GUI and headless live mode.

New Features:
- Introduce Resolution dropdown in the Camera card to choose webcam capture size for live mode.
- Introduce Det size dropdown in the Camera card to configure face detection input resolution and reinitialize the analyser on change.
- Expose a --det-size CLI flag to control face detection input size in headless runs.

Enhancements:
- Make the face analyser read a configurable global det_size and allow forcing a re-init when it changes.
- Have the webcam preview use the global capture_resolution instead of hardcoded preview dimensions.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -59,7 +59,7 @@ def parse_args() -> None:
     program.add_argument('--max-memory', help='maximum amount of RAM in GB', dest='max_memory', type=int, default=suggest_max_memory())
     program.add_argument('--execution-provider', help='execution provider', dest='execution_provider', default=[suggest_default_execution_provider()], choices=suggest_execution_providers(), nargs='+')
     program.add_argument('--execution-threads', help='number of execution threads', dest='execution_threads', type=int, default=suggest_execution_threads())
-    program.add_argument('--det-size', help='face detection input size (160, 320, or 640)', dest='det_size', type=int, default=640, choices=[160, 320, 640])
+    program.add_argument('--det-size', help='face detection input size (160, 320, or 640)', dest='det_size', type=int, default=modules.globals.DEFAULT_DET_SIZE, choices=[160, 320, 640])
     program.add_argument('-v', '--version', action='version', version=f'{modules.metadata.name} {modules.metadata.version}')
 
     # register deprecated args
--- a/modules/core.py
+++ b/modules/core.py
@@ -59,6 +59,7 @@ def parse_args() -> None:
     program.add_argument('--max-memory', help='maximum amount of RAM in GB', dest='max_memory', type=int, default=suggest_max_memory())
     program.add_argument('--execution-provider', help='execution provider', dest='execution_provider', default=[suggest_default_execution_provider()], choices=suggest_execution_providers(), nargs='+')
     program.add_argument('--execution-threads', help='number of execution threads', dest='execution_threads', type=int, default=suggest_execution_threads())
+    program.add_argument('--det-size', help='face detection input size (160, 320, or 640)', dest='det_size', type=int, default=640, choices=[160, 320, 640])
     program.add_argument('-v', '--version', action='version', version=f'{modules.metadata.name} {modules.metadata.version}')
 
     # register deprecated args
@@ -88,6 +89,7 @@ def parse_args() -> None:
     modules.globals.max_memory = args.max_memory
     modules.globals.execution_providers = decode_execution_providers(args.execution_provider)
     modules.globals.execution_threads = args.execution_threads
+    modules.globals.det_size = args.det_size
     modules.globals.lang = args.lang
 
     #for ENHANCER tumblers:
--- a/modules/face_analyser.py
+++ b/modules/face_analyser.py
@@ -54,8 +54,9 @@ def get_face_analyser() -> Any:
                     providers=providers,
                     allowed_modules=['detection', 'recognition', 'landmark_2d_106']
                 )
-                analyser.prepare(ctx_id=0, det_size=_current_det_size())
-                _optimize_det_model(analyser, providers)
+                det_size = _current_det_size()
+                analyser.prepare(ctx_id=0, det_size=det_size)
+                _optimize_det_model(analyser, providers, det_size)
                 # Publish only after prepare()/optimize() complete, so a
                 # lock-free reader can never observe a half-built analyser
                 # (insightface fixes det shape at prepare() time).
@@ -65,7 +66,7 @@ def get_face_analyser() -> Any:
     return analyser
 
 
-def _optimize_det_model(fa: Any, providers) -> None:
+def _optimize_det_model(fa: Any, providers, det_size: tuple) -> None:
     """Replace the detection model's ONNX session with a CoreML-optimized one.
 
     Folds dynamic Shape→Gather chains into constants (the input size is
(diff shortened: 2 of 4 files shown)


5uck1ess:
Rebased onto current `main` (resolves a trivial `import cv2` conflict — the unicode `imread`/`imwrite` refactor made it unused) and addressed the review feedback:

- **Persisted resolution validation** — a zero/negative `capture_resolution` from a stale `switch_states.json` is now rejected before it ever reaches `cv2`, so a bad persisted value can't break capture.
- **det_size threading** — `_optimize_det_model` now takes `det_size` as a parameter instead of re-reading the global, so `prepare()` and the optimize pass can't disagree under a concurrent change.

Ruff gate (`E701,E711,E712,F401,F541`) clean; live-tested preview + persistence on CUDA.

massimomazzariol:
I tested a Windows webcam case that may be relevant to this PR.

This PR exposes capture resolution, which is useful, but on my setup resolution selection alone was not enough because OpenCV/DirectShow silently negotiated the wrong camera format

## Environment

- Windows 11
- Python 3.11.9
- OpenCV 4.10.0
- DirectML / AMD RX 6900 XT
- Webcam: Generic Camera marked as "HD Web Camera"

The webcam supports MJPEG 1280x720@30. Verified outside the app with:

```powershell
ffplay -f dshow -vcodec mjpeg -video_size 1280x720 -framerate 30 -i video="HD Web Camera"
```

## Problem observed

When Deep-Live-Cam requested 1280x720@30 MJPG, OpenCV initially returned YUY2 instead:

```text
requested mode 1280x720@30fps codec=MJPG
actual mode 1280x720; reported_fps=30.0; measured_fps=4.1; fourcc=YUY2
```

So the requested resolution was applied, but the negotiated pixel format was not usable for real-time capture.

## Local fix

The local fix was not just selecting 1280x720. The important part was changing the DirectShow property negotiation order in `VideoCapturer`.

The working strategy was:

1. open camera by DirectShow index
2. set width
3. set height
4. set fps
5. set FOURCC to MJPG
6. read frames and verify actual FOURCC + measured FPS

The relevant local logic looked like this:

```python
cap = cv2.VideoCapture(device_index, cv2.CAP_DSHOW)

cap.set(cv2.CAP_PROP_FRAME_WIDTH, width)
cap.set(cv2.CAP_PROP_FRAME_HEIGHT, height)
cap.set(cv2.CAP_PROP_FPS, fps)

if requested_codec:
    fourcc = cv2.VideoWriter_fourcc(*requested_codec)
    cap.set(cv2.CAP_PROP_FOURCC, fourcc)

# Then verify what the backend actually negotiated.
actual_width = int(cap.get(cv2.CAP_PROP_FRAME_WIDTH))
actual_height = int(cap.get(cv2.CAP_PROP_FRAME_HEIGHT))
reported_fps = cap.get(cv2.CAP_PROP_FPS)
actual_fourcc_int = int(cap.get(cv2.CAP_PROP_FOURCC))

actual_fourcc = "".join(
    chr((actual_fourcc_int >> 8 * i) & 0xFF)
    for i in range(4)
).strip()

# Measure real capture FPS by reading a few frames.
# If requested MJPG silently became YUY2/YUYV at high resolution,
# or measured FPS is far below requested FPS, reject this attempt
# and try another capture strategy/fallback.
```

With this order, the same camera produced:

```text
actual=1280x720; reported_fps=30.0; measured_fps=30.8; fourcc=MJPG
```

## Why this may matter for this PR

This is not a replacement for the resolution dropdown. It is complementary.

The dropdown lets users request the capture size, but the `VideoCapturer` layer should verify that the backend actually negotiated a usable format and frame rate.

Otherwise users may select 720p/1080p and still get a very slow uncompressed YUY2/YUYV stream.

A possible safeguard would be:

```python
if requested_codec == "MJPG" and actual_fourcc in {"YUY2", "YUYV"} and actual_width >= 1280:
    reject_this_capture_strategy()

if requested_fps and measured_fps < requested_fps * 0.5:
    reject_this_capture_strategy()
```

In my local test, the successful strategy was specifically:

```text
DirectShow index / set size first / then set FOURCC
```

and not:

```text
DirectShow index / open params
DirectShow index / set codec first
```

So it may be worth adding actual FOURCC/FPS verification after applying the requested capture resolution.

kier007:
not ready yet, but the resolution controls and saved settings are working, and CI passes, the remaining blocker is the CoreML cache, switching from 640 to 320/160 still reuses the old model and fails inference, please include the input shape in the cache key

5uck1ess:
good catch, fixed in 90df32f. coreml cache is keyed on input size now so switching 640 -> 320/160 rebuilds instead of reusing the old one

kier007 (approved):
aight, lgtm

PR #1835, opened 2026-05-19
web support

hmwl:
## Summary by Sourcery

Add a Flask-based web interface providing browser access to Deep Live Cam for image face swapping and live webcam streaming.

New Features:
- Introduce a single-page HTML Web UI with tabs for live webcam face swapping and still-image face swapping, including parameter controls and visual comparison tools.
- Expose REST APIs for face swapping, random face generation, and live camera management (start/stop, settings, MJPEG stream, and camera enumeration) via a new Flask application module.

Enhancements:
- Configure global processing defaults and model preloading for headless web operation, including live session management and real-time parameter updates without restarting the process.

### DIFF
--- a/web_app.py
+++ b/web_app.py
@@ -1,0 +1,615 @@
+"""
+Deep-Live-Cam Web Interface
+Flask-based web UI for face swapping on images + live webcam streaming.
+"""
+
+import os
+import sys
+import base64
+import logging
+import threading
+import queue
+import time
+import urllib.request
+
+# Ensure project root is on path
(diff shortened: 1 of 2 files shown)


kier007:
Great idea but not applicable for now, the issues are:
- it accepts large base64 images directly in json with no obvious size limit, which it can make memory spikes easier
- the streaming route reads from a global _live session inside the generator, if the user stops and restarts the camera while a stream is open, the old stream can accidentally attach to the new session
- the live ui has an "enhancer" setting but the backend never actually applies it in live mode, so the control is misleading

PR #1836, opened 2026-05-20
fix(ui): Guard preview resize against empty frames

Gujiassh:
## Summary

- move preview image sizing into a small helper that handles missing, empty, or malformed frames safely
- return a black fallback frame for invalid preview input instead of crashing on `.shape` unpacking or zero-sized dimensions
- add focused tests with lightweight fakes so the sizing edge cases can run without GPU/OpenCV dependencies

## Verification

- `PYTHONPATH=. python3 tests/test_image_sizing.py`
- `git diff --check`

## Summary by Sourcery

Improve robustness of image preview sizing by centralizing it in a dedicated helper and adding coverage for edge cases.

Bug Fixes:
- Prevent crashes when preview frames are missing, empty, or malformed by returning a safe fallback frame instead of operating on invalid shapes.

Enhancements:
- Introduce a dedicated _image_sizing module with a reusable fit_image_to_size helper that validates inputs and enforces minimum non-zero dimensions.

Tests:
- Add unit tests for fit_image_to_size using lightweight fakes to cover malformed inputs and sizing edge cases without GPU or OpenCV dependencies.

### DIFF
--- a/modules/_image_sizing.py
+++ b/modules/_image_sizing.py
@@ -1,0 +1,52 @@
+from __future__ import annotations
+
+from typing import Optional, Tuple
+
+import numpy as np
+
+from modules.gpu_processing import gpu_resize
+
+
+def _fallback_frame(width: int, height: int) -> np.ndarray:
+    safe_width = max(1, int(width))
+    safe_height = max(1, int(height))
+    return np.zeros((safe_height, safe_width, 3), dtype=np.uint8)
+
+
+def fit_image_to_size(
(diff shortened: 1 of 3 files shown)


kier007:
Lgtm, but i wouldn't merge this yet. The resize ration can still produce frames larger than the target dimensions, restore the `min(target_wisth / w, target_height / h)` and add a test for mismatched aspect ratios. After that, this should be good to go.

PR #1844, opened 2026-05-23
Fix: resolve 4 confirmed bugs across modules

### LINKED ISSUE #1839: Bug: macOS memory limit uses 1024^6 (exabytes) instead of 1024^3 (gigabytes)
## Description

In \limit_resources()\ in \modules/core.py\, the memory limit calculation for macOS uses \1024 ** 6\ (exabytes) instead of \1024 ** 3\ (gigabytes), making the setrlimit call set an astronomically large value.

## Root Cause

\\\python
def limit_resources() -> None:
    if modules.globals.max_memory:
        memory = modules.globals.max_memory * 1024 ** 3
        if platform.system().lower() == 'darwin':
            memory = modules.globals.max_memory * 1024 ** 6  # BUG: should be 1024 ** 3
        ...
\\\

- \1024 ** 3\ = 1,073,741,824 (≈ 1 GB in bytes) ✓
- \1024 ** 6\ = 1,152,921,504,606,846,976 (≈ 1 EB in bytes) ✗

For \max_memory = 4\:
- Linux/Windows: \4 * 1,073,741,824 = 4,294,967,296\ (≈ 4 GB) ✓
- macOS: \4 * 1,152,921,504,606,846,976 = 4,611,686,018,427,387,904\ (≈ 4 EB) ✗

This appears to be a copy-paste error from the Windows block which uses \ctypes.c_size_t\.

## Expected behavior

macOS memory limit should be calculated the same as the default case: \max_memory * 1024 ** 3\.

## Files affected

- \modules/core.py\ line 175

## Severity

Medium — breaks memory limiting on macOS. The \setrlimit(RLIMIT_DATA, ...)\ call will either fail or set no practical limit.

### LINKED ISSUE #1840: Bug: get_unique_faces_from_target_image/video can crash when get_many_faces() returns None
## Description

\get_unique_faces_from_target_image()\ and \get_unique_faces_from_target_video()\ in \modules/face_analyser.py\ call \get_many_faces()\ which can return \None\ (on \IndexError\), but the callers immediately iterate over the result without a None check.

## Root Cause

\modules/face_analyser.py:154-162\:
\\\python
def get_many_faces(frame: Frame) -> Any:
    try:
        if _is_dml():
            with modules.globals.dml_lock:
                return _analyse_faces(frame)
        else:
            return _analyse_faces(frame)
    except IndexError:
        return None  # <-- Can return None
\\\

\modules/face_analyser.py:258-262\:
\\\python
many_faces = get_many_faces(target_frame)
...
for face in many_faces:  # <-- TypeError when many_faces is None
\\\

Same pattern exists in \get_unique_faces_from_target_video()\ at line 293-295.

## Expected behavior

Check for \None\ before iterating, e.g.:
\\\python
many_faces = get_many_faces(target_frame)
if many_faces is not None:
    for face in many_faces:
        ...
\\\

## Files affected

- \modules/face_analyser.py\ lines 262, 295

## Severity

Medium — crashes the face mapping dialog when no faces are detected in the target.

### LINKED ISSUE #1841: Bug: imwrite_unicode() always returns False for files with extensions (write never happens)
## Description

The utility function `imwrite_unicode()` in `modules/__init__.py` has a logic error that prevents it from writing image files in most cases.

## Root Cause

```python
def imwrite_unicode(path, img, params=None):
    root, ext = os.path.splitext(path)
    if not ext:
        ext = ".png"
        result, encoded_img = cv2.imencode(ext, img, params if params else [])
        result, encoded_img = cv2.imencode(f".{ext}", img, params if params is not None else [])
        encoded_img.tofile(path)
        return True
    return False
```

- `os.path.splitext('output.jpg')` returns `('output', '.jpg')` — ext is `.jpg` (truthy)
- `if not ext:` is `False`, so it skips the write block and hits `return False`
- **For any file with an extension, the function returns False without writing anything**

Additionally for the extensionless case:
- First `cv2.imencode("png", ...)` is missing the required dot prefix (should be `.png`)
- The second `cv2.imencode(f".{ext}", ...)` is correct but the result of the first is discarded

## Expected behavior

The function should write the image to the specified path regardless of whether it has a file extension.

## Files affected

- `modules/__init__.py` lines 10-18

## Severity

Critical — breaks all image writes that go through this utility function.

### LINKED ISSUE #1843: Bug: load_frame_processor_module silently exits when module lacks interface method
## Description

`load_frame_processor_module()` in `modules/processors/frame/core.py` calls `sys.exit()` without printing any error message when a frame processor module is missing a required interface method.

## Root Cause

```python
def load_frame_processor_module(frame_processor: str) -> Any:
    if frame_processor not in ALLOWED_PROCESSORS:
        print(f"Frame processor {frame_processor} is not allowed")
        sys.exit()                          # OK: has error message
    try:
        frame_processor_module = importlib.import_module(...)
        for method_name in FRAME_PROCESSORS_INTERFACE:
            if not hasattr(frame_processor_module, method_name):
                sys.exit()                  # BUG: silent, no error message
    except ImportError:
        print(f"Frame processor {frame_processor} not found")
        sys.exit()                          # OK: has error message
```

When a module is found but doesn't implement one of the required interface methods (`pre_check`, `pre_start`, `process_frame`, `process_image`, `process_video`), the program silently terminates with no user-facing explanation.

## Expected behavior

Print an error message identifying which module is missing which method before exiting.

## Files affected

- `modules/processors/frame/core.py` line 40

## Severity

Low — difficult to debug when it happens.

hklcf:
## Summary

This PR fixes 4 confirmed bugs found during a full code audit.

### 1. Fix imwrite_unicode returning False for files with extensions (Fixes #1841)
**File:** modules/__init__.py

The function had inverted logic: os.path.splitext('file.jpg') returns ext='.jpg' (truthy), so \if not ext:\ was False, causing the function to return False without writing. Also removed duplicate cv2.imencode call.

### 2. Fix macOS memory limit using wrong multiplier (Fixes #1839)
**File:** modules/core.py

1024 ** 6 produced ~4 exabytes instead of ~4 gigabytes. Removed the redundant macOS branch.

### 3. Fix crash when get_many_faces() returns None (Fixes #1840)
**File:** modules/face_analyser.py

get_many_faces() can return None on IndexError, but callers iterated without a None check. Added guards in get_unique_faces_from_target_image() and get_unique_faces_from_target_video().

### 4. Fix silent sys.exit() in load_frame_processor_module (Fixes #1843)
**File:** modules/processors/frame/core.py

Added error message before sys.exit() when a module lacks an interface method.

All 4 bugs were verified with test scripts before making changes.

## Summary by Sourcery

Fix multiple runtime bugs across image writing, face analysis, resource limiting, and frame processor loading.

Bug Fixes:
- Ensure imwrite_unicode correctly writes files with extensions and returns False only when encoding fails.
- Prevent crashes in unique face extraction when get_many_faces returns None for images or video frames.
- Use the correct memory multiplier for macOS by removing the erroneous platform-specific branch that set an excessively large limit.
- Emit a clear error message before exiting when a frame processor module is missing required interface methods.

### DIFF
--- a/modules/__init__.py
+++ b/modules/__init__.py
@@ -11,8 +11,8 @@ def imwrite_unicode(path, img, params=None):
     root, ext = os.path.splitext(path)
     if not ext:
         ext = ".png"
-        result, encoded_img = cv2.imencode(ext, img, params if params else [])
-        result, encoded_img = cv2.imencode(f".{ext}", img, params if params is not None else [])
-        encoded_img.tofile(path)
-        return True
-    return False
+    result, encoded_img = cv2.imencode(ext, img, params if params is not None else [])
+    if not result:
+        return False
+    encoded_img.tofile(path)
+    return True
--- a/modules/core.py
+++ b/modules/core.py
@@ -171,8 +171,6 @@ def limit_resources() -> None:
     # limit memory usage
     if modules.globals.max_memory:
         memory = modules.globals.max_memory * 1024 ** 3
-        if platform.system().lower() == 'darwin':
-            memory = modules.globals.max_memory * 1024 ** 6
         if platform.system().lower() == 'windows':
             import ctypes
             kernel32 = ctypes.windll.kernel32
--- a/modules/face_analyser.py
+++ b/modules/face_analyser.py
@@ -257,6 +257,8 @@ def get_unique_faces_from_target_image() -> Any:
         modules.globals.source_target_map = []
         target_frame = cv2.imread(modules.globals.target_path)
         many_faces = get_many_faces(target_frame)
+        if many_faces is None:
+            return None
         i = 0
 
         for face in many_faces:
@@ -291,6 +293,8 @@ def get_unique_faces_from_target_video() -> Any:
         for temp_frame_path in tqdm(temp_frame_paths, desc="Extracting face embeddings from frames"):
             temp_frame = cv2.imread(temp_frame_path)
             many_faces = get_many_faces(temp_frame)
+            if many_faces is None:
+                continue
 
             for face in many_faces:
                 face_embeddings.append(face.normed_embedding)
--- a/modules/processors/frame/core.py
+++ b/modules/processors/frame/core.py
@@ -37,6 +37,7 @@ def load_frame_processor_module(frame_processor: str) -> Any:
         frame_processor_module = importlib.import_module(f'modules.processors.frame.{frame_processor}')
         for method_name in FRAME_PROCESSORS_INTERFACE:
             if not hasattr(frame_processor_module, method_name):
+                print(f"Frame processor {frame_processor} is missing required method {method_name}")
                 sys.exit()
     except ImportError:
         print(f"Frame processor {frame_processor} not found")


kier007:
lgtm

PR #1846, opened 2026-05-23
feat: virtual camera output (pyvirtualcam → OBS Virtual Camera)

5uck1ess:
## Summary

Adds a **Virtual Cam** toggle in the Options card. When enabled at Live start, processed frames are sent to a virtual webcam (OBS Virtual Camera driver on Windows/macOS, `v4l2loopback` on Linux) so apps like Discord, Zoom, Google Meet, and Teams can pick the swapped face up as a webcam input.

Output is fixed at **1280×720 / 30fps** so meeting apps see a stable size regardless of the underlying capture resolution.

## Why

Currently the swapped output is preview-window only — useful for demos but not for the workflow most people actually want (looking like someone else on a video call). `pyvirtualcam` is the de-facto Python bridge to the OBS Virtual Camera driver (Windows + macOS) and `v4l2loopback` (Linux). Users still need to install OBS Studio for the driver itself; this is documented in the tooltip and the failure-path status messages.

## Why fixed at 1280×720 / 30fps (and not configurable)

- **Meeting-app standard.** Zoom, Meet, Teams, Discord all default to 720p and many cap higher inputs to 720p anyway. Going to 1080p costs noticeable extra CPU/GPU for the upscale+send without visible quality gain on the receiving end.
- **Aspect stability.** 16:9 is what every meeting app expects, and `_aspect_crop_to` already handles the common 4:3 (e.g. 640×480) negotiated-input case so faces aren't horizontally squashed.
- **30fps ceiling.** Meeting apps generally cap at 30fps; higher just costs more compute for no perceived smoothness gain.
- **pyvirtualcam constraint.** The vcam stream's width/height/fps are set at `pyvirtualcam.Camera()` creation and can't change mid-session, so any configurability would need to be a one-time choice at Live-start (e.g. a dropdown) rather than a live setting.

If/when #1833 (capture resolution + det_size dropdowns) lands, the vcam output dimensions could be tied to the user-selected capture resolution — but that PR isn't in yet, and upstream main currently has no user-facing resolution config to hook into. Exposing `VCAM_W`/`VCAM_H`/`VCAM_FPS` as user-configurable is intentionally kept out of scope here to keep this PR minimal and focused on the vcam plumbing itself. Happy to do it as a separate PR once #1833 (or some equivalent resolution selector) is in.

## What changed

- **`requirements.txt`** — `pyvirtualcam>=0.15.0` (with a comment about the OBS Studio host requirement)
- **`modules/globals.py`** — `virtual_cam: bool = False` toggle
- **`modules/ui.py`**:
  - `VCAM_W=1280`, `VCAM_H=720`, `VCAM_FPS=30` constants
  - `_aspect_crop_to(frame, w, h)` — channel-count-agnostic center crop to the target aspect before resize. Many webcams negotiate 4:3 (e.g. 640×480) even when 16:9 is requested, and a bare resize to 1280×720 horizontally stretches faces.
  - `WebcamPreviewWindow.__init__` — opens `pyvirtualcam.Camera()` when `virtual_cam` is True; clear status messages on `ImportError` / driver-missing without killing the preview loop.
  - When vcam is active the preview window shrinks to ~360×140 and shows a static status message; closing it via X stops capture / processing / vcam — same UX as a normal preview close.
  - `_tick` sends each processed frame to vcam **before** the display fit-resize, with wall-clock pacing (`time.monotonic()` gating, not `cam.sleep_until_next_frame()` because we're inside a `QTimer` callback and can't block the Qt event loop) to honor the advertised 30fps even when the webcam path delivers frames faster.
  - On send failure the vcam handle is closed before being dropped, a single status message is emitted via `update_status()`, and the preview loop continues without vcam.
  - `virtual_cam` is included in `save_switch_states` / `load_switch_states` so the toggle persists across launches like the other switches.

## Tested

Verified on Windows + Google Meet with OBS Virtual Camera installed — swapped face shows up at 720p in the call.

## Re #1834

Re-submission of #1834 (closed) with @kier007's review feedback applied:

1. Explicit frame pacing to honor the 30fps contract
2. `self._vcam.close()` on send failure before disabling, so the handle doesn't leak until preview close
3. Toggle persistence via `save_switch_states` / `load_switch_states`
4. Status-channel logging (`update_status()`) instead of `print()` on send error

Also includes the @sourcery-ai fix from that PR (drop explicit 3rd-axis slice in `_aspect_crop_to` for channel-count robustness).

### DIFF
--- a/modules/globals.py
+++ b/modules/globals.py
@@ -70,6 +70,13 @@ enable_interpolation: bool = True # Toggle temporal smoothing
 interpolation_weight: float = 0  # Blend weight for current frame (0.0-1.0). Lower=smoother.
 # --- END: Added for Frame Interpolation ---
 
+# Virtual camera output (pyvirtualcam → OBS Virtual Camera driver).
+# When enabled, processed frames are upscaled to VCAM 1280x720 and sent to
+# the virtual cam so apps like Discord/Meet/Zoom can pick it up as a
+# webcam. Requires OBS Studio installed (for the driver). Toggle is
+# read at Live-start; mid-Live toggling currently requires Stop+Start.
+virtual_cam: bool = False
+
 # --- END OF FILE globals.py ---
 
 import threading
--- a/modules/ui.py
+++ b/modules/ui.py
@@ -610,8 +610,6 @@ class MainWindow(QMainWindow):
 
             def _on_toggle(v, f=field):
                 setattr(modules.globals, f, v)
-                if f == "virtual_cam":
-                    print(f"[vcam] toggle -> globals.virtual_cam = {v}")
                 save_switch_states()
 
             sw.toggled.connect(_on_toggle)
@@ -1216,8 +1214,9 @@ class _ProcessingWorker(QThread):
 
 
 class WebcamPreviewWindow(QWidget):
-    def __init__(self, camera_index: int):
+    def __init__(self, camera_index: int, on_close: Optional[Callable[[], None]] = None):
         super().__init__()
+        self._on_close_callback = on_close
         self.setWindowTitle("Live Preview")
         self.resize(PREVIEW_DEFAULT_WIDTH, PREVIEW_DEFAULT_HEIGHT)
         layout = QVBoxLayout(self)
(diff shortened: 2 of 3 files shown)


kier007:
Request Change:

- PR body says that linux uses v4l2loopback but the UI error says to install OBS studio for the driver, it's fine for windows/macOS, but misleading for Linux, fix platform-specific for virtual camera error/help text, make sure to document the host driver setup in README, especially Linux v4l2loopback

PR #1847, opened 2026-05-23
Bump onnxruntime to 1.26.0, drop Pad(reflect) CoreML workaround

maxwbuckley:
## Summary

- Bump `onnxruntime-silicon==1.16.3` → stock `onnxruntime==1.26.0` on Apple Silicon, and `onnxruntime-gpu==1.23.2` → `1.26.0` on other platforms.
- Delete the `_decompose_reflect_pad` pass (`modules/onnx_optimize.py`, –163 / +10 lines) and its dead `_rp_` initializer filter in `_preserve_emap_position`.

## Why this is safe

ORT 1.26.0 (shipped 2026-05-08) lands [microsoft/onnxruntime#28073](https://github.com/microsoft/onnxruntime/pull/28073), which adds native MLProgram support for `Pad(mode=reflect)` on the CoreML EP. The Slice+Concat decomposition we were applying to `inswapper_128.onnx` (14 `Pad(reflect)` ops) is no longer required.

Verified empirically with ORT verbose logging:

```
[V:onnxruntime:, session_state.cc:1350 VerifyEachNodeIsAssignedToAnEp]
  All nodes placed on [CoreMLExecutionProvider]. Number of nodes: 1
```

The whole graph fuses into a single CoreML partition — zero CPU fallback for any `Pad(reflect)` node, no graph breaks.

## Why the silicon fork goes away

`onnxruntime-silicon` on PyPI is effectively unmaintained — last release was 1.16.3 (Oct 2023), it never shipped a 1.26 wheel, and Python 3.13 wheels are not available. Stock `onnxruntime` ships CoreML support natively on Apple Silicon, so the fork is no longer needed for ANE acceleration. As a bonus this unblocks Python 3.13 on macOS.

## Perf evidence (Apple Silicon, inswapper_128, mean ± noise across a short session, std-dev not captured)

| ORT | `_decompose_reflect_pad` | Approx FPS |
|---|---|---|
| `onnxruntime-silicon==1.16.3` | applied | <6 |
| `onnxruntime==1.24.4` | applied | 21–25 |
| `onnxruntime==1.24.4` | removed | ~12 |
| `onnxruntime==1.26.0` | removed (this PR) | ~30 |

Notes on the matrix:

- The silicon 1.16.3 baseline is two years of CoreML EP improvements behind — that explains the <6 FPS, not anything about the model.
- On 1.24.4, removing the decomposition halves FPS, confirming 1.24's CoreML EP still graph-breaks on `Pad(reflect)`. So the hack should not be removed against pre-1.26 ORT.
- On 1.26.0, removing the decomposition recovers ~30 FPS — i.e. the native MLProgram Pad(reflect) path is on par with (or faster than) the manual Slice+Concat rewrite, and the model still runs as a single CoreML partition.

## Code

The three remaining passes in `modules/onnx_optimize.py` — Shape/Gather folding, Split → Slice decomposition, scalar-Gather widening — are unchanged.

## Test plan

- [ ] Run `python run.py` on Apple Silicon (Python 3.11 or 3.13) and confirm `inswapper_128` swap works end-to-end.
- [ ] Run on a CUDA box and confirm `onnxruntime-gpu==1.26.0` still loads cleanly.
- [ ] Confirm no regression on `face_enhancer` (GFPGAN) — that model never used the reflect-pad pass.

## Summary by Sourcery

Upgrade ONNX Runtime dependencies to 1.26.0 and remove the now-unnecessary Pad(reflect) CoreML optimization pass while keeping remaining ONNX optimization passes intact.

Enhancements:
- Drop the custom Pad(reflect) → Slice+Concat graph rewrite now that CoreML supports Pad(mode=reflect) natively in ONNX Runtime 1.26.0.
- Simplify CoreML optimization pass numbering and clean up the embedding map initializer selection logic after removing Pad(reflect)-related artifacts.

Build:
- Update Python dependencies to use stock onnxruntime 1.26.0 on Apple Silicon and onnxruntime-gpu 1.26.0 on non-macOS platforms, replacing the unmaintained onnxruntime-silicon fork.

### DIFF
--- a/modules/onnx_optimize.py
+++ b/modules/onnx_optimize.py
@@ -9,17 +9,11 @@ would otherwise introduce:
    inference with the known input size and replace these chains with constants.
    Float32-noise-level differences only (max ~6e-6).
 
-2. **Pad(reflect) decomposition** — CoreML doesn't support ``Pad(mode=reflect)``.
-   Models using reflect padding (e.g. inswapper_128) get split into many CoreML
-   subgraphs with CPU fallbacks between each.  We rewrite each ``Pad(reflect)``
-   as equivalent ``Slice`` + ``Concat`` ops that CoreML handles natively.
-   Bit-for-bit identical output. (Fixed upstream in microsoft/onnxruntime#28073.)
-
-3. **Split → Slice decomposition** — CoreML's EP doesn't support the ONNX
+2. **Split → Slice decomposition** — CoreML's EP doesn't support the ONNX
    ``Split`` op, causing partition boundaries in models with channel-wise
    splits (e.g. GFPGAN's SFT modulation). Each 2-way Split becomes two Slices.
 
-4. **Scalar Gather widening** — ORT's CoreML EP rejects ``Gather`` nodes with
+3. **Scalar Gather widening** — ORT's CoreML EP rejects ``Gather`` nodes with
    rank-0 (scalar) indices. StyleGAN-derived models (GFPGAN) slice per-layer
    style codes using exactly this pattern. We widen each scalar index to
    ``[1]`` and squeeze the added axis on the Gather output.
@@ -67,10 +61,6 @@ def optimize_for_coreml(model_path: str, input_shape: tuple = None) -> str:
     changed = False
 
     if _fold_shape_gather(model, input_shape):
-        changed = True
-
-    # TODO(ort>=1.26): drop this pass. Fixed upstream by microsoft/onnxruntime#28073.
-    if _decompose_reflect_pad(model):
         changed = True
 
     if _decompose_split(model):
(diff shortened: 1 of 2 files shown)


kier007:
In the line 51 of onnx_optimize.py, it returns the *_coreml.onnx immediately when it is newer than the source model. So anyone who already generated inswapper_128_coreml.onnx with the old Pad(reflect) rewrite will keep loading the stale Slice/Concat model after this PR, esp from line 248 face_swapper.py. It should invalidate/version the cache or use a new suffix so ORT 1.26 actually gets the original native Pad(reflect) graph.

The PR is good idea, mostly, but I would not merge as-is until the cache invalidation issue is handled. The dependency rationale checks out the ONNX Runtime 1.26.0 release note list CoreML Pad reflect support/fixes, and PyPl has onxruntime / onnxruntime-gpu 1.26.0 packages.

Sources:
https://github.com/microsoft/onnxruntime/releases
https://pypi.org/project/onnxruntime/1.26.0/
https://pypi.org/project/onnxruntime-gpu/1.26.0/

PR #1849, opened 2026-05-24
fix: skip setrlimit(RLIMIT_DATA) on macOS

### LINKED ISSUE #1848: macOS: app crashes on launch with "current limit exceeds maximum limit"
On macOS, launching the app crashes immediately:

```
File "modules/core.py", line 180, in limit_resources
ValueError: current limit exceeds maximum limit
```

The macOS kernel rejects `setrlimit(RLIMIT_DATA, ...)` with `EINVAL` for the values used here; CPython's `resource` module reports any `EINVAL` from `setrlimit` as the misleading `ValueError` above.

PR to follow.

iikuzmychov:
Fixes #1848.

The macOS kernel rejects `setrlimit(RLIMIT_DATA, ...)` with EINVAL for the values used here, crashing the app on launch. Skip the call on darwin.

## Summary by Sourcery

Bug Fixes:
- Prevent crashes on macOS by not calling setrlimit(RLIMIT_DATA) with unsupported values when limiting memory usage.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -170,8 +170,8 @@ def limit_resources() -> None:
             tensorflow.config.experimental.set_memory_growth(gpu, True)
     # limit memory usage
     if modules.globals.max_memory:
-        # setrlimit(RLIMIT_DATA) is unusable on macOS.
-        # See https://github.com/python/cpython/issues/142317
+        # setrlimit(RLIMIT_DATA) fails with EINVAL on macOS, crashing on launch.
+        # See https://github.com/hacksider/Deep-Live-Cam/issues/1848
         if platform.system().lower() == 'darwin':
             return
         memory = modules.globals.max_memory * 1024 ** 3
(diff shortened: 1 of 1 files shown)


noahximus:
One small correction before merging: python/cpython#142317 concerns RLIMIT_STACK rather than RLIMIT_DATA. Please remove that reference or replace it with a link to Deep-Live-Cam issue #1848. The platform guard itself looks appropriate for the reported startup failure.

iikuzmychov:
@noahximus you're right, updated

noahximus:
Sorry for the long silence — you made this change back in July and it sat. Thanks for the correction, and for tracking down the original cause in #1848.

Merging as-is.

For the record on why the early return rather than a try/except: `--max-memory` defaults to `suggest_max_memory()`, which returns `4` on macOS, so `modules.globals.max_memory` is always truthy there and every launch reaches the `setrlimit` call. Catching the error would work, but it would print a warning on every single macOS start for a condition the platform can't satisfy. Skipping outright is the better default.

PR #1863, opened 2026-06-24
feat: GFPGAN model selector with hot-swap

5uck1ess:
Adds a GFPGAN model selector with live hot-swap. Re-submission of #1832 (closed); same branch, rebuilt against current `main` with the earlier review feedback applied.

## What
- UI selector to switch the GFPGAN ONNX variant (1024 high-quality ↔ 512 lighter/faster) without restart.

## Review feedback addressed
- **Crash-safe load** — `get_face_enhancer` validates and returns the session inside `THREAD_LOCK`; a failed load sets `FACE_ENHANCER = None` and raises instead of leaving a half-initialized global.
- **No Qt deadlock** — stopped calling `update_status` while holding the non-reentrant lock (Qt's `processEvents` re-entrancy could deadlock it); model load now logs to console.
- **Variant persists** — switching model updates `gfpgan_model_filename` so the choice survives restart; switching away to a GPEN variant or None drops the cached GFPGAN session to free VRAM.
- **Cache race fixed** — `enhance_face` snapshots the temporal cache under the lock so a concurrent `reset_face_enhancer()` can't hand `_paste_back` a half-cleared matrix.

Ruff gate (`E701,E711,E712,F401,F541`) clean; live-tested 1024↔512 hot-swap and a full GPEN round-trip; variant persisted across restart.

## Summary by Sourcery

Add selectable GFPGAN ONNX variants with hot-swappable model switching and safer session management in the face enhancer pipeline.

New Features:
- Introduce a UI dropdown to choose between 1024 and 512 GFPGAN ONNX variants for face enhancement.
- Persist the selected GFPGAN model filename across sessions via switch state storage.

Enhancements:
- Make GFPGAN model loading crash-safe and thread-safe, including validation under a shared lock and clearer fallback behavior when models are missing or invalid.
- Guard and reset internal face enhancement caches when switching models to avoid mismatched outputs and races.
- Normalize persisted face processor UI state to ensure all expected keys are present when loading.

### DIFF
--- a/modules/globals.py
+++ b/modules/globals.py
@@ -70,6 +70,13 @@ enable_interpolation: bool = True # Toggle temporal smoothing
 interpolation_weight: float = 0  # Blend weight for current frame (0.0-1.0). Lower=smoother.
 # --- END: Added for Frame Interpolation ---
 
+# GFPGAN model filename. Currently shipped variants:
+#   gfpgan-1024.onnx  — 512 in, 1024 out (super-res head). Highest quality.
+#   GFPGANv1.4.onnx   — 512 in, 512 out. Lighter, faster.
+# Both files are hosted on the hacksider/deep-live-cam HF model repo.
+# Hot-swap supported via reset_face_enhancer() (called from UI on change).
+gfpgan_model_filename: str = "gfpgan-1024.onnx"
+
 # --- END OF FILE globals.py ---
 
 import threading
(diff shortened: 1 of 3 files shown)


kier007:
Nice improvement over #1832 — the missing-model path, persisted selection, and session/mask reset issues look much better now.

I think there’s still one small hot-swap race before this is ready: `_enh_live_cache` is cleared under `THREAD_LOCK`, but it gets repopulated after inference without taking that lock. If a frame using the old session finishes just after `reset_face_enhancer()`, it can put the old model’s output back into the cache and the next frame may reuse it. Writing those cache fields under the same lock, or tagging the work with a model generation and dropping stale results, should close the gap.

Other than that, the scope is clean and the targeted Ruff checks pass. Pretty close.

5uck1ess:
I just can't get it perfect huh? fixed. Hopefully that's almost there :)

PR #1864, opened 2026-06-24
feat: WEBP source image support

5uck1ess:
Adds WEBP as a supported source-image format. Re-submission of #1831 (closed); branch is the same, rebuilt against current `main` with the earlier review feedback applied.

## What
- WEBP source images now load and swap correctly, decoded via the libwebp bundled with opencv-python.

## Review feedback addressed
- **GIF dropped** — OpenCV's `imread`/`imwrite` can't decode or encode GIF on 4.10/4.11, so offering it would silently fail; it's excluded instead.
- **Centralized extension lists** — `IMAGE_EXTENSIONS`/`VIDEO_EXTENSIONS` are defined once in `globals.py`; the video save-dialog filter is built from `VIDEO_EXTENSIONS` (`_VIDEO_FILE_FILTER`) so the dialogs and `has_image_extension` can't drift.
- Removed the dead `file_types` list (verified unused in both fork and upstream).

Ruff gate (`E701,E711,E712,F401,F541`) clean; live-tested load + swap.

## Summary by Sourcery

Add WEBP as a supported image format and centralize media file extension handling across globals and UI dialogs.

New Features:
- Support loading WEBP images as valid source images by recognizing their extensions and integrating with existing image detection logic.

Enhancements:
- Centralize canonical image and video extension lists in globals and reuse them to build QFileDialog filters, keeping UI dialogs and extension checks in sync.
- Remove the unused file_types definition in globals to simplify media type configuration.

### DIFF
--- a/modules/globals.py
+++ b/modules/globals.py
@@ -7,7 +7,7 @@ ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
 WORKFLOW_DIR = os.path.join(ROOT_DIR, "workflow")
 
 file_types = [
-    ("Image", ("*.png", "*.jpg", "*.jpeg", "*.gif", "*.bmp")),
+    ("Image", ("*.png", "*.jpg", "*.jpeg", "*.gif", "*.bmp", "*.webp")),
     ("Video", ("*.mp4", "*.mkv")),
 ]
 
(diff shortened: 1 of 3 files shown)


kier007:
lgtm

PR #1869, opened 2026-06-26
fix: release CUDA tensors in _fast_paste_back to prevent GPU memory exhaustion

### LINKED ISSUE #1868: GPU memory leak in _fast_paste_back causes CUBLAS_STATUS_NOT_INITIALIZED crash on long videos
## Description

When processing long videos (>500 frames) with `--execution-provider cuda`, Deep-Live-Cam crashes partway through with:

- `CUBLAS_STATUS_NOT_INITIALIZED`
- `CUDA failure 715: illegal instruction`

The crash is caused by a GPU memory leak in `_fast_paste_back` in `modules/processors/frame/face_swapper.py`.

## Root Cause

`_fast_paste_back` has a PyTorch CUDA path that creates intermediate tensors on the GPU but never explicitly frees them:

```python
mask_t = torch.from_numpy(mask_crop).float().cuda()
fake_t = torch.from_numpy(bgr_fake_crop).float().cuda()
tgt_t = torch.from_numpy(target_crop).float().cuda()
blended = (mask_t * fake_t + (1.0 - mask_t) * tgt_t).to(torch.uint8).cpu().numpy()
```

`mask_t`, `fake_t`, `tgt_t`, and `blended` persist in GPU memory until Python's GC runs. ONNX Runtime (insightface) also holds GPU memory on the same device. Over hundreds of frames, the unreleased PyTorch tensors accumulate, exhausting VRAM and corrupting the CUDA context.

## Reproduction

1. Select a source face image and a target video with ~1000+ frames at 1080p
2. Run: `python run.py -s face.jpg -t test.mp4 -o output.mp4 --execution-provider cuda`
3. Observe crash partway through (typically after ~300-500 frames on a 6GB GPU)

## Environment

- GPU: NVIDIA RTX 2060 6GB
- Driver: 596.21
- CUDA: 12.4
- OS: Windows 11
- Python: 3.11.14
- onnxruntime-gpu: 1.18.0
- PyTorch: 2.6.0+cu124

## Fix

1. Add `del mask_t, fake_t, tgt_t, blended` after the blend operation to immediately release tensor references
2. Add periodic `gc.collect()` + `torch.cuda.empty_cache()` every ~50 frames in `_run_pipe_pipeline`
3. Make `release_resources()` attempt `torch.cuda.empty_cache()` via runtime import instead of requiring a module-level `HAS_TORCH` flag

I have a verified working fix (benchmarked: 1129 frames completed without crash, ~8.9 fps) and can open a PR.

caiyilian:
Fixes #1868

## Summary

GPU memory leak in _fast_paste_back causes CUBLAS_STATUS_NOT_INITIALIZED crash when processing long videos (>500 frames) with --execution-provider cuda.

## Root Cause

_fast_paste_back creates PyTorch CUDA intermediate tensors (mask_t, ake_t, 	gt_t, lended) that are never explicitly freed. ONNX Runtime (insightface) and PyTorch share the same GPU. Over hundreds of frames, the unreleased tensors accumulate, exhausting VRAM and corrupting the CUDA context.

### modules/processors/frame/face_swapper.py
- Add del mask_t, fake_t, tgt_t, blended after blend to immediately release CUDA tensor references

### modules/processors/frame/core.py
- Add gc.collect() + 	orch.cuda.empty_cache() every 50 frames in _run_pipe_pipeline

### modules/core.py
- Make 
elease_resources() attempt 	orch.cuda.empty_cache() via runtime import instead of module-level HAS_TORCH flag

## Benchmark

Tested on RTX 2060 6GB, 1129 frames (1080p, 37.6s video):
- **Before fix:** Crash at ~300-500 frames with CUBLAS error
- **After fix:** Completed 1129 frames in 126.8s (~8.9 fps) without errors

## Summary by Sourcery

Address GPU memory exhaustion during CUDA-based frame processing by releasing intermediate tensors and proactively clearing CUDA caches.

Bug Fixes:
- Free intermediate CUDA tensors after fast paste-back blending to prevent VRAM accumulation and CUDA context corruption.
- Periodically trigger Python garbage collection and CUDA cache clearing during frame pipeline processing to reduce long-run memory growth.
- Ensure resource release attempts CUDA cache clearing via a safe runtime import, even when PyTorch was not imported at module load time.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -59,6 +59,7 @@ def parse_args() -> None:
     program.add_argument('--max-memory', help='maximum amount of RAM in GB', dest='max_memory', type=int, default=suggest_max_memory())
     program.add_argument('--execution-provider', help='execution provider', dest='execution_provider', default=[suggest_default_execution_provider()], choices=suggest_execution_providers(), nargs='+')
     program.add_argument('--execution-threads', help='number of execution threads', dest='execution_threads', type=int, default=suggest_execution_threads())
+    program.add_argument('--cache-clean-interval', help='frames between CUDA cache flushes (0 to disable)', dest='cache_clean_interval', type=int, default=50)
     program.add_argument('-v', '--version', action='version', version=f'{modules.metadata.name} {modules.metadata.version}')
 
     # register deprecated args
@@ -88,6 +89,7 @@ def parse_args() -> None:
     modules.globals.max_memory = args.max_memory
     modules.globals.execution_providers = decode_execution_providers(args.execution_provider)
     modules.globals.execution_threads = args.execution_threads
+    modules.globals.cache_clean_interval = max(0, args.cache_clean_interval)
     modules.globals.lang = args.lang
 
     #for ENHANCER tumblers:
--- a/modules/core.py
+++ b/modules/core.py
@@ -182,6 +182,10 @@ def limit_resources() -> None:
 
 def release_resources() -> None:
     if 'CUDAExecutionProvider' in modules.globals.execution_providers:
+        # Import torch at runtime rather than relying on a module-level
+        # HAS_TORCH flag. This ensures the CUDA cache is flushed even when
+        # PyTorch is loaded after this module, and avoids import-order issues
+        # that could prevent cleanup on shutdown (issue #1868).
         try:
             import torch
             torch.cuda.empty_cache()
--- a/modules/core.py
+++ b/modules/core.py
@@ -188,7 +188,8 @@ def release_resources() -> None:
         # that could prevent cleanup on shutdown (issue #1868).
         try:
             import torch
-            torch.cuda.empty_cache()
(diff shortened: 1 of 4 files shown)


kier007:
niceee but this looks a reasonable mitigation and the RTX 2060 repro is useful, one thing i'd tighten before merging: only import torch and run `gc.collect()`/`empty_cache()` when `CUDAExecutionProvider` is actually active, right now CPU/CoreML runs pay that cost every 50 frames too, i'd also soften the "tensor leak" wording a bit - the tensors are locals and should die when `_fast_paste_back()` returns, so the actual win here is probably returning PyTorch's cached blocks to CUDA so ONNX Runtime can use them. A quick before/after log of `memory_allocated()` vs `memory_reserved()` would make that story much stronger, with that and CI green, this lgtm :D

PR #1870, opened 2026-06-26
feat: auto-skip face_swapper when no source face is provided

### LINKED ISSUE #1867: Feature Request: Skip face swap when only target file is provided and face enhancement is enabled
### Summary

Allow the application to process a target image/video using only face enhancement (e.g., GFPGAN, GPEN) **without requiring a source face image**, when the user explicitly enables a face enhancer. Currently, uploading only a target file (no source) results in a silent no-op in the UI, or crashes in headless mode.

### Current Behavior

1. **UI (Start button):** When only a target is uploaded and the user clicks Start, `_select_output_and_start()` silently returns (`modules/ui.py` lines 885-898) because `is_image(None)` and `is_video(None)` both return `False`. No status message is shown to the user.

2. **Headless / CLI (`run.py -t target.mp4`):** `has_image_extension(None)` raises an `AttributeError` (`modules/utilities.py` line 265) because `None.lower()` is called.

3. **Even when face enhancement is enabled:** There is no code path that skips the swap step and runs only the enhancer — the face swapper always runs first and crashes or produces no output when `source_face` is `None`.

### Why This Matters

The face enhancers (GFPGAN, GPEN-256, GPEN-512) are **entirely source-independent** — they detect and enhance faces directly on the target frame without needing a source face. You can verify this in the current code:

- `face_enhancer.py:process_frame()` — accepts `source_face` but **never uses it**
- `face_enhancer.py:enhance_face()` — detects faces via `get_many_faces(temp_frame)` and enhances them directly
- Same pattern in `face_enhancer_gpen256.py` and `face_enhancer_gpen512.py`

This makes it a natural standalone "face restoration" mode.

### Proposed Behavior

This effectively gives users a **face restoration / upscaling tool** without needing a source image — a common use case for enhancing old photos or low-quality video.

### Suggested Implementation Areas

1. **`modules/ui.py` — `_on_start()` / `_select_output_and_start()`:** 
   - When `target_path` is set but `source_path` is `None`, check if any enhancer is active in `fp_ui`.
   - If yes, skip the output file dialog for swap and proceed to run only enhancement processors.
   - If no enhancer is active, show the error message above.

2. **`modules/core.py` — `start()`:**
   - Before calling `frame_processor.process_image()`, check if `source_path` is `None`.
   - If `source_path` is `None` and only enhancers are configured (no swapper), skip the swap and copy steps, running only enhancement.

3. **`modules/utilities.py` — `has_image_extension()` / `is_image()` / `is_video()`:**
   - Add `None` guards to prevent `AttributeError` when paths are `None`.

4. **`modules/processors/frame/core.py` — processor pipeline:**
   - When `source_path` is `None`, filter out the face swapper from the processor pipeline, keeping only enhancers.

### Additional Considerations

- **Live / webcam mode** is already exempt (it already works independently as long as a source is provided), so this change should only affect image/video file processing.
- The "map_faces" mode should also get the corresponding guard.
- A status message like *"Processing with face enhancement only (no face swap)..."* would help users understand what is happening.

### Related Context

- The current silent return in `_select_output_and_start()` when `target_path` is falsy makes it hard for users to understand why nothing happens.
- This was discovered during code analysis of the target-only and source-only upload paths.

caiyilian:
When running with no source image (-s) but an enhancer enabled, the face swapper has nothing to swap. This PR auto-detects this case and removes face_swapper from the pipeline, allowing the enhancer to run in face-restoration mode.

Fixes #1867

## Summary by Sourcery

Improve GPU memory handling during frame processing and refine CUDA cache cleanup behavior across the pipeline.

Enhancements:
- Add periodic Python GC and CUDA cache flushing during in-memory video processing to reduce VRAM accumulation for long runs.
- Release temporary CUDA tensors explicitly after face blending to avoid per-frame VRAM growth when using the face swapper.
- Change resource release logic to import PyTorch at runtime and conditionally flush the CUDA cache, avoiding import-order issues.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -206,6 +206,16 @@ def start() -> None:
     
     start_time = time.time()
     
+    # When no source is provided but an enhancer is active, automatically
+    # remove the face_swapper and run in face-restoration mode (#1867).
+    if modules.globals.source_path is None:
+        enhancers = {'face_enhancer', 'face_enhancer_gpen256', 'face_enhancer_gpen512'}
+        if enhancers & set(modules.globals.frame_processors):
+            modules.globals.frame_processors = [
+                p for p in modules.globals.frame_processors if p != 'face_swapper'
+            ]
+            update_status('No source face — running in face-restoration mode (enhancement only).')
+    
     for frame_processor in get_frame_processors_modules(modules.globals.frame_processors):
         if not frame_processor.pre_start():
             return


kier007:
check my comment in #1869

PR #1872, opened 2026-06-28
fix: resolve OpenVINO DLL loading on Windows for OpenVINOExecutionProvider

dunegym:
## Description

Fixes the `openvino.dll` loading issue that prevented ONNX Runtime's OpenVINOExecutionProvider from working on Windows.

### Root Cause

`onnxruntime_providers_openvino.dll` depends on `openvino.dll`, but the OpenVINO Python package installs `openvino.dll` under `site-packages/openvino/libs/` — a directory not on the system `PATH`. When ONNX Runtime attempts to create a session with `OpenVINOExecutionProvider`, the DLL load fails silently and falls back to CPU.

### Changes

| File | Change |
|------|--------|
| `run.py` | Call `add_openvino_libs_to_path()` from `onnxruntime.tools.add_openvino_win_libs` before any ONNX session is created (graceful no-op when OpenVINO is absent) |
| `modules/platform_info.py` | Detect `OpenVINOExecutionProvider` in available providers; show "OpenVINO (Intel)" in the banner |
| `modules/core.py` | Add `openvino` to provider priority list (`cuda > rocm > coreml > openvino > dml > cpu`); set thread hint to 1 for OpenVINO EP |
| `modules/processors/frame/_onnx_enhancer.py` | Configure `OpenVINOExecutionProvider` with `device_type: GPU, precision: FP16` for best performance |
| `modules/processors/frame/face_swapper.py` | Same GPU FP16 provider config applied to the face swap model |

### Performance

`inswapper_128.onnx` benchmark on Intel GPU + OpenVINO:

| Provider | Device | Latency | FPS |
|----------|--------|---------|-----|
| OpenVINO EP | GPU FP16 | **76 ms** | **13.1 FPS** |
| CPU EP (fallback) | CPU | 786 ms | 1.3 FPS |

### Testing

- [x] `--execution-provider openvino` listed in `--help`
- [x] Banner shows `accelerator: OpenVINO (Intel)`
- [x] Face swapper model loads with `OpenVINOExecutionProvider` active
- [x] Inference produces correct output shape `(1, 3, 128, 128)`
- [x] Registration is a no-op when OpenVINO is not installed (try/except)

## Summary by Sourcery

Ensure OpenVINOExecutionProvider works and is preferred when available, particularly on Windows with Intel GPUs.

New Features:
- Detect and label OpenVINOExecutionProvider in platform info, showing it as "OpenVINO (Intel)" in the accelerator banner.

Bug Fixes:
- Register OpenVINO DLL directories on Windows at startup so onnxruntime can successfully load OpenVINOExecutionProvider instead of silently falling back to CPU.

Enhancements:
- Prioritize OpenVINO ahead of DirectML in the default execution provider selection and set a single-thread hint when OpenVINOExecutionProvider is active.
- Configure OpenVINOExecutionProvider with GPU FP16 settings for ONNX-based frame enhancement and face swapping models to improve performance.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -132,9 +132,9 @@ def suggest_max_memory() -> int:
 
 
 def suggest_default_execution_provider() -> str:
-    """Pick the best available provider: cuda > rocm > coreml > dml > cpu."""
+    """Pick the best available provider: cuda > rocm > coreml > openvino > dml > cpu."""
     available = encode_execution_providers(onnxruntime.get_available_providers())
-    for pref in ('cuda', 'rocm', 'coreml', 'dml'):
+    for pref in ('cuda', 'rocm', 'coreml', 'openvino', 'dml'):
         if pref in available:
             return pref
     return 'cpu'
@@ -157,6 +157,8 @@ def suggest_execution_threads() -> int:
         return 1
     if 'CUDAExecutionProvider' in modules.globals.execution_providers:
         return 2
+    if 'OpenVINOExecutionProvider' in modules.globals.execution_providers:
+        return 1
     
     # For CPU execution, use most cores but leave some for system
     return max(4, min(cpu_count - 2, 16))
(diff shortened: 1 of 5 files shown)


kier007:
nice direction overall, but i think this needs a couple of fixes before merging:
1) `add_openvino_libs_to_path()` can call `sys.exit()` when the OpenVINO libraries aren't installed, and the current `except` won't catch `SystemExit`, that means some windows setups could exit during startup instead of gracefully falling back
2) the provider config always forces `GPU` + `FP16`, having `OpenVINOExecutionProvider` available doesn't necessarily mean the machine has a supported Intel GPU-OpenVINO also supports CPU and NPU, it would be safer to use the default/`AUTO`, make the device configurable or add a fallback
3) the OpenVINO thread hint currently won't take effect because `suggest_execution_threads()` runs before `modules.globals.execution_providers` is assigned

the DLL- path fix itself make sense; these edge cases just need tightening up first

PR #1875, opened 2026-07-02
Claude/great wright 52yygf

vranky:
## Summary by Sourcery

Introduce a higher-resolution pixel-boost face swap path while keeping the existing 128x128 swap as a fallback.

New Features:
- Add a pixel-boosted face swap mode that performs swaps on higher-resolution, pixel-interleaved crops and reassembles them into a high-res face.

Enhancements:
- Guard face swapping with DML lock for both standard and pixel-boosted paths and reuse _fast_paste_back with dynamically sized alpha masks based on the generated swap output.

### DIFF
--- a/modules/processors/frame/face_swapper.py
+++ b/modules/processors/frame/face_swapper.py
@@ -229,16 +229,30 @@ def get_face_swapper() -> Any:
 
     with THREAD_LOCK:
         if FACE_SWAPPER is None:
-            # Prefer FP16 on GPUs with Tensor Cores (Turing+) — half the
-            # memory bandwidth, faster inference.  Fall back to FP32 for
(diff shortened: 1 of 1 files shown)


kier007:
hey, that's a cool idea, i saw the pixel interleaving approach looks good, a few things i'd probably tighten up before merging are:
1) the `PIXEL_BOOST = 2` is always on, so every face now needs 4 model runs per frame, that could hit live FPS pretty hard, especially with multiple faces, would be nice to make it optional and keep 1 as the default
2) the `emap` check only warns and continues loading, if it's invalid, swapping can fail every frame and just return the original image, probably better to fail early or fall back to the FP32 model
3) the mask still uses the same blur at 256x256, so the feather becomes narrower and might create a visible seams, scaling the blur with the boost should preserve the old blending

PR #1878, opened 2026-07-10
fix: scope PyTorch CUDA cache cleanup to CUDA pipelines

### LINKED ISSUE #1868: GPU memory leak in _fast_paste_back causes CUBLAS_STATUS_NOT_INITIALIZED crash on long videos

caiyilian:
Fixes #1868

Supersedes #1869, which GitHub no longer permits reopening.

## What changed
- Load PyTorch, run cleanup, and release its CUDA cache only when CUDAExecutionProvider is active. CPU and CoreML pipelines do not pay this cost.
- Release unused PyTorch cache blocks every 50 frames. Full cyclic garbage collection runs every 500 frames, reducing pauses while retaining periodic cleanup.
- Report memory_allocated() and memory_reserved() before and after each cache release through the existing DLC.CORE status channel. If CUDA memory statistics or cache release fails, processing continues safely.
- Use the same CUDA provider constant for pipeline and shutdown cleanup.

## Validation
- Added 9 focused CUDA-memory tests plus the existing 3 tests: 12 unit tests pass.
- End-to-end CUDA run on RTX 3060 12GB with PyTorch 2.6.0+cu124 and ONNX Runtime CUDAExecutionProvider: completed all 1,129 frames of the supplied 1080x1920, 37.63-second video. Total runtime: 82.50 seconds; in-memory processing and encoding: 80.31 seconds.
- The status logs show allocated memory at 0 MiB before and after cleanup, while reserved cache fell from 88 MiB to 0 MiB at frame 50 and from 258 MiB to 0 MiB at frame 1,100. This confirms the benefit is returning cached blocks to the CUDA driver rather than retaining live local tensors.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -26,7 +26,7 @@ except ImportError:
 import modules.globals
 import modules.metadata
 import modules.ui as ui
-from modules.processors.frame.core import get_frame_processors_modules, process_video_in_memory
+from modules.processors.frame.core import CUDA_EXECUTION_PROVIDER, get_frame_processors_modules, process_video_in_memory
 from modules.utilities import has_image_extension, is_image, is_video, detect_fps, create_video, extract_frames, get_temp_frame_paths, restore_audio, create_temp, move_temp, clean_temp, normalize_output_path
 
 if HAS_TORCH and 'ROCMExecutionProvider' in modules.globals.execution_providers:
@@ -181,8 +181,16 @@ def limit_resources() -> None:
 
 
 def release_resources() -> None:
-    if 'CUDAExecutionProvider' in modules.globals.execution_providers and HAS_TORCH:
-        torch.cuda.empty_cache()
+    if CUDA_EXECUTION_PROVIDER in modules.globals.execution_providers:
+        # Avoid the module-level availability snapshot: PyTorch may be loaded
+        # after this module and still have CUDA cache blocks to release.
+        try:
+            import torch
+        except ImportError:
+            return
+
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
 
 
 def pre_check() -> bool:
--- a/modules/processors/frame/core.py
+++ b/modules/processors/frame/core.py
@@ -1,6 +1,7 @@
 import os
 import subprocess
 import sys
+import gc
 import importlib
 from concurrent.futures import ThreadPoolExecutor
 from types import ModuleType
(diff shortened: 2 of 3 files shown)


kier007:
This is a solid improvement over #1869, but I’m a little hesitant about calling torch.cuda.empty_cache() every 50 frames for every CUDA run. That interval feels fairly arbitrary and could hurt throughput by forcing PyTorch to reallocate memory it would normally reuse—especially on GPUs with plenty of VRAM.
The RTX 3060 result proves the stability benefit, but a same-machine A/B benchmark with cleanup enabled and disabled would help quantify the performance cost. It would also be nice to make the interval configurable or trigger cleanup only when reserved memory crosses a threshold. With that, an update from main, and green CI, I’d be comfortable merging this.

PR #1879, opened 2026-07-11
fix: resolve OpenVINO DLL loading on Windows for OpenVINOExecutionProvider

dunegym:
## Description

Fixes the `openvino.dll` loading issue that prevented ONNX Runtime's OpenVINOExecutionProvider from working on Windows.

### Root Cause

`onnxruntime_providers_openvino.dll` depends on `openvino.dll`, but the OpenVINO Python package installs `openvino.dll` under `site-packages/openvino/libs/` — a directory not on the system `PATH`. When ONNX Runtime attempts to create a session with `OpenVINOExecutionProvider`, the DLL load fails silently and falls back to CPU.

### Changes

| File | Change |
|------|--------|
| `run.py` | Call `add_openvino_libs_to_path()` from `onnxruntime.tools.add_openvino_win_libs` before any ONNX session is created (graceful no-op when OpenVINO is absent, catches SystemExit) |
| `modules/platform_info.py` | Detect `OpenVINOExecutionProvider` in available providers; show "OpenVINO (Intel)" in the banner |
| `modules/core.py` | Add `openvino` to provider priority list; set thread hint to 1 for OpenVINO EP; defer thread evaluation until after providers are assigned |
| `modules/processors/frame/_onnx_enhancer.py` | Extract shared `OPENVINO_PROVIDER_CONFIG` with `AUTO:GPU,NPU,CPU` device priority; used by both enhancer and face_swapper |
| `modules/processors/frame/face_swapper.py` | Import shared OpenVINO config constant |

### PR Review Fixes

This revision addresses feedback from collaborator `kier007` on the original PR:

1. **SystemExit** — `add_openvino_libs_to_path()` can call `sys.exit()` when OpenVINO is absent; now caught alongside `ImportError`/ `FileNotFoundError`
2. **Hard-coded GPU+FP16** — Changed to `AUTO:GPU,NPU,CPU` so OpenVINO selects the best device available; no longer assumes an Intel GPU is present
3. **Thread timing bug** — `suggest_execution_threads()` was evaluated at module load time before `execution_providers` was assigned; now deferred until after `parse_args()` finishes

### Performance

`inswapper_128.onnx` on Intel Arc iGPU + OpenVINO:

| Provider | Device | Latency | FPS |
|----------|--------|---------|-----|
| OpenVINO EP | AUTO:GPU,NPU,CPU | **78 ms** | **12.8 FPS** |
| CPU EP (fallback) | CPU | 786 ms | 1.3 FPS |

### Testing

- [x] `--execution-provider openvino` listed in `--help`
- [x] Banner shows `accelerator: OpenVINO (Intel)`
- [x] Face swapper model loads with `OpenVINOExecutionProvider` active
- [x] Thread hint correctly returns 1 for OpenVINO EP
- [x] Registration is a no-op when OpenVINO is absent (catches SystemExit)
- [x] Removing openvino package gracefully falls back to CPU

## Summary by Sourcery

Handle OpenVINO DLL registration and provider configuration so OpenVINOExecutionProvider works reliably, especially on Windows, and integrate it into provider selection, threading, and UI accelerator labeling.

New Features:
- Add OpenVINOExecutionProvider to the default execution provider priority list and expose it as a selectable accelerator option.
- Introduce a shared OpenVINO provider configuration that uses AUTO:GPU,NPU,CPU for device selection in both the ONNX enhancer and face swapper.

Bug Fixes:
- Ensure suggest_execution_threads() is evaluated only after execution providers are assigned to avoid incorrect thread recommendations.
- Register OpenVINO DLL paths on Windows at startup so onnxruntime can successfully load openvino.dll and avoid silent fallback to CPU.
- Treat missing OpenVINO installation as a no-op by catching ImportError, FileNotFoundError, and SystemExit during DLL registration.

Enhancements:
- Detect OpenVINOExecutionProvider in available ONNX providers and label it as "OpenVINO (Intel)" in the accelerator banner.
- Assign a thread hint of 1 when OpenVINOExecutionProvider is active to better match its performance characteristics.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -58,7 +58,7 @@ def parse_args() -> None:
     program.add_argument('--live-resizable', help='The live camera frame is resizable', dest='live_resizable', action='store_true', default=False)
     program.add_argument('--max-memory', help='maximum amount of RAM in GB', dest='max_memory', type=int, default=suggest_max_memory())
     program.add_argument('--execution-provider', help='execution provider', dest='execution_provider', default=[suggest_default_execution_provider()], choices=suggest_execution_providers(), nargs='+')
-    program.add_argument('--execution-threads', help='number of execution threads', dest='execution_threads', type=int, default=suggest_execution_threads())
+    program.add_argument('--execution-threads', help='number of execution threads', dest='execution_threads', type=int, default=None)
     program.add_argument('-v', '--version', action='version', version=f'{modules.metadata.name} {modules.metadata.version}')
 
     # register deprecated args
@@ -89,6 +89,12 @@ def parse_args() -> None:
     modules.globals.execution_providers = decode_execution_providers(args.execution_provider)
     modules.globals.execution_threads = args.execution_threads
     modules.globals.lang = args.lang
+
+    # The argparse default (None) avoids evaluating suggest_execution_threads()
+    # before providers are decoded, and deprecated-arg overrides above may
+    # have already set execution_threads.
+    if modules.globals.execution_threads is None:
+        modules.globals.execution_threads = suggest_execution_threads()
 
     #for ENHANCER tumblers:
     for enhancer_key in ('face_enhancer', 'face_enhancer_gpen256', 'face_enhancer_gpen512'):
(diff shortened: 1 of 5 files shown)


kier007:
lgtm

PR #1884, opened 2026-07-20
fix: improve macOS support for memory suggestion and video capture

antazo:
This change adjusts the camera initialization path to use platform-specific OpenCV backends and configuration parameters, improving compatibility and reducing capture failures on supported systems.

## Summary by Sourcery

Improve macOS compatibility for video capture and memory usage suggestions.

Bug Fixes:
- Adjust macOS camera initialization to prefer the AVFoundation backend and avoid failing with incompatible backends.
- Update macOS max memory suggestion to skip RLIMIT_DATA adjustments that cannot be safely lowered.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -132,8 +132,9 @@ def decode_execution_providers(execution_providers: List[str]) -> List[str]:
 
 
 def suggest_max_memory() -> int:
+    # macOS RLIMIT_DATA cannot be lowered to a finite value; skip by default.
     if platform.system().lower() == 'darwin':
-        return 4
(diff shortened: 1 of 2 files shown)


antazo:
Thanks for the feedback. I’ve updated the change to keep the platform-specific behavior focused and explicit: the macOS camera path now prefers AVFoundation and logs backend attempts for debugging, and the memory-suggestion change now treats the macOS special case explicitly so it doesn’t assume a positive finite value. This keeps the patch aligned with the original intent while making the behavior clearer and easier to maintain.

PR #1885, opened 2026-07-20
fix: replace unverified SSL context; make shell=False explicit in utilities.py

anupamme:
## Summary

Two changes to `modules/utilities.py`:

1. **SSL fix (real improvement):** Replace `ssl._create_unverified_context()` on macOS with `ssl.create_default_context()` on all platforms. The old code disabled TLS certificate verification for all model downloads on macOS, leaving downloads open to MITM. The new code uses Python's default verified context everywhere.

2. **subprocess hardening (defensive clarity):** Add explicit `shell=False` to `subprocess.check_output` in `run_ffmpeg`. This does not change runtime behavior — `commands` was already a list and `shell=False` is the default — but makes the intent visible to future readers and static analyzers.

## What this is not

The original PR description claimed a CWE-78 command injection. That was a scanner false positive: `commands` was already constructed as a list and passed directly to `subprocess.check_output` without `shell=True`, so there was no shell interpretation occurring. The explicit `shell=False` adds clarity, not safety.

## Changes
- `modules/utilities.py`: replace macOS unverified SSL context with `ssl.create_default_context()`; add explicit `shell=False`; remove unused `platform` import (fixes ruff F401 CI failure)

## Verification
- `python -m py_compile modules/utilities.py` passes
- `ruff check modules/utilities.py` passes (no F401)
- Existing test suite passes
- Download flow on macOS: **not yet tested** — flagged by reviewer, worth checking before merge

### DIFF
--- a/modules/utilities.py
+++ b/modules/utilities.py
@@ -28,7 +28,7 @@ def run_ffmpeg(args: List[str]) -> bool:
     ]
     commands.extend(args)
     try:
-        subprocess.check_output(commands, stderr=subprocess.STDOUT)
+        subprocess.check_output(commands, stderr=subprocess.STDOUT, shell=False)
         return True
     except subprocess.CalledProcessError as error:
         output = error.output.decode(errors="ignore").strip()
@@ -290,10 +290,7 @@ def conditional_download(download_directory_path: str, urls: List[str]) -> None:
             request = urllib.request.Request(url)
             
             # Create a specific SSL context for macOS to avoid globally disabling verification
-            ctx = None
-            if platform.system().lower() == "darwin":
-                ctx = ssl._create_unverified_context()
-                
+            ctx = ssl.create_default_context()
             response = urllib.request.urlopen(request, context=ctx)
             total = int(response.headers.get("Content-Length", 0))
             with tqdm(
--- a/modules/utilities.py
+++ b/modules/utilities.py
@@ -288,10 +288,9 @@ def conditional_download(download_directory_path: str, urls: List[str]) -> None:
         )
         if not os.path.exists(download_file_path):
             request = urllib.request.Request(url)
-            
-            # Create a specific SSL context for macOS to avoid globally disabling verification
-            ctx = ssl.create_default_context()
-            response = urllib.request.urlopen(request, context=ctx)
+
+            # Use default SSL context for secure certificate verification
+            response = urllib.request.urlopen(request, context=ssl.create_default_context())
             total = int(response.headers.get("Content-Length", 0))
             with tqdm(
                 total=total,
--- a/modules/utilities.py
+++ b/modules/utilities.py
@@ -289,7 +289,6 @@ def conditional_download(download_directory_path: str, urls: List[str]) -> None:
         if not os.path.exists(download_file_path):
             request = urllib.request.Request(url)
 
-            # Use default SSL context for secure certificate verification
             response = urllib.request.urlopen(request, context=ssl.create_default_context())
             total = int(response.headers.get("Content-Length", 0))
             with tqdm(
--- a/modules/utilities.py
+++ b/modules/utilities.py
@@ -1,7 +1,6 @@
 import glob
 import mimetypes
 import os
-import platform
 import shutil
 import ssl
 import subprocess


anupamme:
> Hey - I've left some high level feedback:
> 
> * The `conditional_download` change now uses `ssl.create_default_context()` on all platforms, but the existing comment still refers to a macOS-specific unverified context; consider updating or removing the comment so it accurately describes the new behavior.
> * In `conditional_download`, since you're always creating a default SSL context, you can simplify the code by inlining `ssl.create_default_context()` directly in the `urlopen` call rather than using an intermediate `ctx` variable.
> 
> Prompt for AI Agents
> ```
> Please address the comments from this code review:
> 
> ## Overall Comments
> - The `conditional_download` change now uses `ssl.create_default_context()` on all platforms, but the existing comment still refers to a macOS-specific unverified context; consider updating or removing the comment so it accurately describes the new behavior.
> - In `conditional_download`, since you're always creating a default SSL context, you can simplify the code by inlining `ssl.create_default_context()` directly in the `urlopen` call rather than using an intermediate `ctx` variable.
> ```
> 
> Sourcery is free for open source - if you like our reviews please consider sharing them ✨
> * [X](https://twitter.com/intent/tweet?text=I%20just%20got%20an%20instant%20code%20review%20from%20%40SourceryAI%2C%20and%20it%20was%20brilliant%21%20It%27s%20free%20for%20open%20source%20and%20has%20a%20free%20trial%20for%20private%20code.%20Check%20it%20out%20https%3A//sourcery.ai/%3Futm_source%3Dtwitter%26utm_medium%3Dsocial%26utm_campaign%3Dbot_review_share)
> * [Mastodon](https://mastodon.social/share?text=I%20just%20got%20an%20instant%20code%20review%20from%20%40SourceryAI%2C%20and%20it%20was%20brilliant%21%20It%27s%20free%20for%20open%20source%20and%20has%20a%20free%20trial%20for%20private%20code.%20Check%20it%20out%20https%3A//sourcery.ai/%3Futm_source%3Dmastodon%26utm_medium%3Dsocial%26utm_campaign%3Dbot_review_share)
> * [LinkedIn](https://www.linkedin.com/sharing/share-offsite/?url=https%3A//sourcery.ai/%3Futm_source%3Dlinkedin%26utm_medium%3Dsocial%26utm_campaign%3Dbot_review_share)
> * [Facebook](https://www.facebook.com/sharer/sharer.php?u=https%3A//sourcery.ai/%3Futm_source%3Dfacebook%26utm_medium%3Dsocial%26utm_campaign%3Dbot_review_share)
> 
> Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.

addressed. pls review.

anupamme:
✅ **Review Feedback Addressed**

I've automatically addressed 1 review comment(s):

The reviewers (sourcery-ai bot) provided two actionable suggestions:

1. **Remove/update the outdated comment** in `conditional_download` - The comment still references macOS-specific unverified context behavior, but the new code uses `ssl.create_default_context()` on all platforms. The comment should be removed or updated to accurately describe the new behavior.

2. **Inline `ssl.create_default_context()`** directly in the `urlopen` call rather than assigning it to a variable first, since it's always used and there's no conditional logic anymore.

The current diff already shows the comment was updated to "# Use default SSL context for secure certificate verification" and `ssl.create_default_context()` is inlined in the `urlopen` call. However, looking at the diff more carefully, the comment is still present. The reviewer wants the comment removed entirely OR the code simplified by inlining without the comment variable assignment.

Looking at the current diff, the code already inlines `ssl.create_default_context()` directly in the `urlopen` call. The remaining issue is the comment "# Use default SSL context for secure certificate verification" - the reviewer says to remove the comment about macOS-specific context. The current diff already replaced the old comment with a new accurate one.

Wait - re-reading the feedback: the reviewer says the comment "still refers to a macOS-specific unverified context" - but in the current diff, the comment was already updated to "# Use default SSL context for secure certificate verification". So the comment issue may already be addressed in the current diff.

The second point about inlining is also already done in the current diff.

Since the user says "Make the required changes" referring to the bot's suggestions, and the current diff already addresses both points, I need to look at what the actual file looks like on the branch to see if there's still work to do. The diff shows the changes that were made - the comment was updated and the code was inlined. 

The remaining actionable item from the reviewer is to remove the comment entirely (since it's now redundant/obvious) to keep the code clean. Let me implement that simplification.

**Files modified:**
- `modules/utilities.py`

The changes have been pushed to this PR branch. Please review!

kier007:
hey @anupamme , i took a look and tried to reproduce the reported command injection, i tested:
- compared the pr base (`ab64c186`) with the current head (`1f79971`) and reviewed the full `modules/utilities.py` file
- also searched for every `run_ffmpeg()` call and traced the arguments through `extract_frames()`, `create_video()`, and `restore_audio()`
- checked for `shell=True`, I didn't find any, the arguments stay as a list all the way into `subprocess.check_ouput()`
- passed a harmless payload containing `;`, `&`, `|`, and `&&` through the same list based subprocess pattern, the whole payload arrived as one literal argument and nothing after the shell characters was executed
- confirmed that `subprocess.Popen` defaults to `shell=False`
- ran `python -m py_compile modules/utilities.py` (it's passed)
- ran `python -m unittest discover -s tests -v` (so all 3 existing tests passed, those tests are mostly for face detection behavior, though, so they don't directly cover this subprocess or TLS change
- ran a security focuses diff scan and manually reviewed the changed code and its callers (no reportable security findings)
- also checked the ruff workflow (it currently fails with `F401` because `platform` is imported but no longer used

so based on those checks, i couldn't reproduce CWE-78, the old code already passed an argument list to `subprocess.check_output(), and `shell=False` is already the default, so adding it explicitly doesn't change the execution behavior, it looks like a scanner false positive rather than a critical command injection fix, the TLS still looks useful because it removes the unverified SSL context, i didn't test the download flow on macOS, though, so that would be worth checking before merge

anupamme:
Hey - thanks for the thorough look; appreciate you actually tracing the call sites and testing payloads rather than just taking the scanner's word for it.

You're right: commands was already a list going into subprocess.check_output, and shell=False is the default, so the explicit flag doesn't change runtime behaviour. This looks like a scanner false positive on my end rather than a real command injection. I'll update the title/description to stop claiming a CVE-style fix and reframe this as a defensive/explicit hardening change (making the "no shell" intent visible in code, which has some value for future readers/auditors even if it's not fixing an active bug).

Given that, I'll:

Drop the CWE-78/"critical vulnerability" framing from the PR description.
Fix the ruff F401 failure by removing the now-unused platform import.
Keep the SSL context change, since that one is a real improvement (removes the unverified-context bypass on macOS).
Test the download flow on macOS before this gets merged, since you flagged that gap.

PR #1896, opened 2026-08-03
fix: sanitize subprocess call in core.py

anupamme:
## Summary
Fix high severity security issue in `modules/processors/frame/core.py`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | HIGH |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `modules/processors/frame/core.py:306` |
| **Assessment** | Likely exploitable |

**Description**: The video processing pipeline spawns FFmpeg subprocesses without resource limits or timeouts, allowing attackers to exhaust system resources with specially crafted video files.

## Evidence

**Exploitation scenario**: Provide a video file with extreme resolution (8K+), infinite duration, or malformed data that causes unbounded resource consumption in FFmpeg processes.

**Scanner confirmation**: multi_agent_ai rule `V-001` flagged this pattern.

**Production code**: This file is in the production codebase, not test-only code.

## Threat Model Context

This is a desktop application - exploitation requires the user to open a crafted file or interact with malicious content.

## Changes
- `modules/processors/frame/core.py`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

## Summary by Sourcery

Tighten frame processor loading and add timeouts to FFmpeg subprocess handling to mitigate resource exhaustion risks.

Bug Fixes:
- Prevent unbounded FFmpeg subprocess execution by adding timeouts and handling timeout expiry during pipeline shutdown.

Enhancements:
- Constrain frame processor imports to a predefined, validated set of modules to avoid loading unexpected processors.

### DIFF
--- a/modules/processors/frame/core.py
+++ b/modules/processors/frame/core.py
@@ -26,7 +26,14 @@ ALLOWED_PROCESSORS = {
     'face_swapper',
     'face_enhancer',
     'face_enhancer_gpen256',
-    'face_enhancer_gpen512'
+    'face_enhancer_gpen512',
+}
+
+_PROCESSOR_MODULES = {
+    'face_swapper': 'modules.processors.frame.face_swapper',
+    'face_enhancer': 'modules.processors.frame.face_enhancer',
+    'face_enhancer_gpen256': 'modules.processors.frame.face_enhancer_gpen256',
+    'face_enhancer_gpen512': 'modules.processors.frame.face_enhancer_gpen512',
 }
 
 def load_frame_processor_module(frame_processor: str) -> Any:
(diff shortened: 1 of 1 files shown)


kier007:
looks like the timeout is being added a bit too late here, it only starts after the read/process/write loop, so if FFmpeg get stucks on `read()` or `write()`, we never reach `wait(timeout=3600)`, might be better to put a timeout around the whole pipeline and kill both process when it fires, also `importlib` and `_PROCESS_MODULES` look unused now, and `importlib` will probably trip Ruff's F401

anupamme:
Thanks for the review. I agree with the concern around the timeout placement. In the current implementation, the existing blocking writer.wait() / reader.wait() calls occur before the new timed waits, so a hung FFmpeg process can prevent execution from ever reaching the timeout. That means the current patch does not actually provide the intended bound.

If you're open to reopening, I can revise the PR to:

* remove the unconditional wait() calls;
* use the timeout directly on the process waits;
* explicitly terminate/kill and reap both FFmpeg processes on timeout;
* add a regression test that exercises a process which doesn’t exit within the timeout;
* clean up the processor-loading change / unused _PROCESSOR_MODULES so the security fix remains focused.

I also agree that a one-hour timeout is not a particularly strong mitigation for resource exhaustion. I can revisit the timeout value and, where appropriate, add input/resource constraints rather than presenting the timeout alone as a complete DoS mitigation.

PR #1902, opened 2026-08-09
fix: no cam detected on Arch Linux, use string path for camera capture instead

1ceseismic:
fix failure to detect and open camera on linux capture device. 
Use explicit `/dev/videoN` paths instead.
Is a rebased follow up to #1783. it's just @Vito-M's original fix onto `bdeeb3a`


kier007:
lgtm

PR #1903, opened 2026-08-09
Add direct virtual camera output on Linux

1ceseismic:
adds optional v4l2 output for processed live frames using a non-blocking writer. UI toggle & setup script + test.
I wanted this instead of having to rely on OBS virtual cam (or similar). much nicer solution

## Summary by Sourcery

Add optional Linux-only virtual camera output for live processed frames and document how to set it up.

New Features:
- Expose a Virtual camera toggle in the live camera UI on Linux to publish processed live output as a webcam via v4l2loopback.
- Introduce a VirtualCameraSink abstraction to stream processed frames to a v4l2loopback-backed device using pyvirtualcam.
- Add a setup script and kernel/module configuration for creating a labeled v4l2loopback device on Linux.

Enhancements:
- Refine live preview frame queue handling to always keep and forward the latest processed frame, including for virtual camera output.

Build:
- Add a Linux-only dependency on pyvirtualcam for virtual camera support.

Deployment:
- Install v4l2loopback configuration files and module-load settings via a helper script to provision the virtual camera device.

Documentation:
- Document Linux virtual camera usage and setup steps in the README.

Tests:
- Add unit tests for virtual camera device discovery and frame format handling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,18 @@ python run.py --execution-provider openvino
 -   Use a screen capture tool like OBS to stream.
 -   To change the face, select a new source image.
 
+### Direct Virtual Camera on Linux
+
+Install and configure `v4l2loopback`:
+
+```bash
+./setup-virtual-camera-linux.sh
+```
+
+Enable **Virtual camera**, select the input camera, and click **Live**. Select
+**Deep Live Cam** in the receiving application. The virtual camera is available
+while the Live Preview is running.
+
 ## Download all models in this huggingface link
 - [**Download models here**](https://huggingface.co/hacksider/deep-live-cam/tree/main)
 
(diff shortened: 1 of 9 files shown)


kier007:
good pr, the queueing and linux dep setup looks clean, gonna hold off on merging just yet, though, the main issue is that if the virtual camera fails to start, the worker exists but the toggle and saved state stay enabled, so the UI says it's on when noting is running, we should reset the toggle/global state when startup fails

1ceseismic:
@kier007 agree, thanks for pointing out. Have now fixed in 2c25c511: a toggle failure now resets the toggle & global/saved state. Failure reason also appears in UI.

1ceseismic:
bump :3

PR #1915, opened 2026-08-22
Improve memory limit handling in limit_resources function

KhademOHAli1:
## Summary by Sourcery

Bug Fixes:
- Make memory limit configuration respect existing hard limits and tolerate platforms that reject finite data limits, preventing startup failures when resource limiting cannot be applied.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -185,7 +185,20 @@ def limit_resources() -> None:
             kernel32.SetProcessWorkingSetSize(-1, ctypes.c_size_t(memory), ctypes.c_size_t(memory))
         else:
             import resource
-            resource.setrlimit(resource.RLIMIT_DATA, (memory, memory))
+            _, hard_limit = resource.getrlimit(resource.RLIMIT_DATA)
+            soft_limit = memory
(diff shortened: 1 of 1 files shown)


noahximus:
Thanks for this — the hard-limit clamping is a real improvement and I'd like to take it.

One change needed first: #1849 just landed, which returns early from `limit_resources()` on Darwin. That was the right call for macOS specifically — `--max-memory` defaults to `suggest_max_memory()` (4 on Darwin), so `max_memory` is always truthy there, and a try/except would print a warning on every single macOS launch for something the platform can't do.

That makes the macOS half of this PR dead code, but the Linux half still stands on its own: clamping the soft limit to the hard limit prevents a `ValueError` when `--max-memory` exceeds the inherited hard limit, which is a genuine bug on Linux today.

Could you rebase on current `main` and scope this to the non-Darwin path — the `getrlimit` clamp plus the `try`/`except` — dropping the macOS-specific comment? The trailing-newline fix at the end of the file is welcome, keep that. Happy to merge quickly once it's rebased.

PR #1924, opened 2026-09-03
Improve swap quality: color matching + exposed frame smoothing

Jeremmya:
## Summary
- Wires a previously-written-but-never-called LAB color-transfer helper into `swap_face()` behind a new **Color Match** toggle, so the swapped face's skin tone/lighting can be matched to the target scene (masked to the face ellipse so hair/background don't skew the stats).
- Exposes the existing but UI-inaccessible temporal interpolation as a **Smoothing** slider (next to Sharpness) to reduce flicker/jitter in video/live mode.
- Fixes `apply_post_processing()` always doing a full-frame `.copy()` even when sharpening/smoothing are both no-ops (the default), by folding the interpolation-weight check into the early-exit condition.

## Why
`inswapper_128` bakes in the source face's skin tone/lighting, which is one of the most common "looks fake" cues when it doesn't match the target scene — Color Match addresses that directly. The smoothing control already existed in the swap pipeline but had no way to be turned on from the UI, so it was dead by default.

## Test plan
- [x] `python3 -m py_compile` on all changed files
- [x] Unit-verified `_match_face_color` pulls the fake face's LAB stats toward a synthetic target's stats
- [x] Unit-verified `cv2.meanStdDev(..., mask=...)` masked-stat usage
- [ ] Manual UI smoke test (Color Match toggle + Smoothing slider) on a live webcam session

## Summary by Sourcery

Improve face-swap realism and temporal stability by adding target-scene color matching and user-configurable smoothing.

New Features:
- Add an optional Color Match control to align swapped face skin tone and lighting with the target scene.
- Expose temporal smoothing through a UI slider for reducing flicker and jitter in video and live modes.

Bug Fixes:
- Avoid unnecessary frame copies when post-processing is effectively inactive.

### DIFF
--- a/modules/globals.py
+++ b/modules/globals.py
@@ -32,6 +32,7 @@ many_faces: bool = False         # Process all detected faces with default sourc
 map_faces: bool = False          # Use source_target_map or simple_map for specific swaps
 poisson_blend: bool = False      # Enable Poisson Blending for smoother face swaps
 color_correction: bool = False   # Enable color correction (implementation specific)
+color_match: bool = False        # Match swapped face's skin tone/lighting to the target scene
 nsfw_filter: bool = False
 
 # Video Output Options
--- a/modules/processors/frame/face_swapper.py
+++ b/modules/processors/frame/face_swapper.py
@@ -65,6 +65,36 @@ def _create_elliptical_mask(size: Tuple[int, int]) -> np.ndarray:
     return mask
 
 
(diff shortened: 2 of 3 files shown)


kier007:
not ready yet, smoothing shares PREVIOUS_FRAME_RESULT across video export workers and can blend frames out of order, please apply it in frame order and add a regression test

PR #1938, opened 2026-09-17
fix: stop live-preview workers on quit, repair random-face fetch

ozp3:
## Problem

The app aborts with `SIGABRT` when you quit while a live preview is open:

```
[webcam] Camera running at 1280x720@30fps
QThread: Destroyed while thread '' is still running
zsh: abort      python run.py
```

There is no Python traceback because nothing raises — Qt calls `qFatal()` itself.

## Root cause

`core.destroy()` ends in a bare `quit()` ([modules/core.py#L337](https://github.com/hacksider/Deep-Live-Cam/blob/main/modules/core.py#L337)). That is Python's builtin `quit()`, so it raises `SystemExit` from inside a Qt slot: the interpreter tears down without unwinding Qt, and `WebcamPreviewWindow`'s `_CaptureWorker` / `_ProcessingWorker` threads are still running when their `QThread` wrappers are destroyed. Destroying a running `QThread` is fatal in Qt.

Five further defects sit on the same code paths and are fixed together, since each one on its own can also reach the fatal case.

## Changes

**1. Stop live-preview workers before the interpreter goes down**
The destroy callback now closes every window that owns worker threads before `destroy()` runs, so the threads finish normally. This is the fix for the abort above.

**2. `closeEvent` gave workers 2s and silently ignored the timeout**
```python
worker.wait(2000)   # returns False on timeout, never raises
except Exception:   # so this branch never fired
```
A first-time GFPGAN load on CoreML takes several seconds, well past 2s, after which a still-running thread could be garbage collected. The timeout is now 30s, and any worker that still has not finished is parked in `_STRAY_WORKERS` rather than destroyed.

**3. `closeEvent` raised `AttributeError` when `__init__` bailed out early**
```
AttributeError: 'WebcamPreviewWindow' object has no attribute '_stop_event'
```
`__init__` returns early when the camera fails to open — which is the normal first-launch path on macOS, before camera permission has been granted. Attribute access in `closeEvent` is now guarded.

**4. The random-face button saved HTML as `.jpg`**
`thispersondoesnotexist.com` no longer serves the image at `/`; that path now returns an HTML landing page, so `raise_for_status()` passes and 4.5 KB of HTML gets written to `deep_live_cam_random_face.jpg`. The image now lives at `/random-person.jpeg`. The response `content-type` is checked as well.

**5. `_on_random_face` assigned `source_path` before validating the download**
`modules.globals.source_path` was set, and only then was the file rendered — so a failed fetch left `source_path` pointing at unreadable data while the error was merely printed.

**6. `imread_unicode()` returning `None` crashed the processing thread**
```
File "insightface/model_zoo/retinaface.py", line 211, in detect
    im_ratio = float(img.shape[0]) / img.shape[1]
AttributeError: 'NoneType' object has no attribute 'shape'
```
Combined with #4 and #5, clicking the random-face button and then Live reliably killed the app. Both call sites (live preview and still preview) now check and report via `update_status` instead.

## Reproduction and validation

The abort reproduces deterministically without a camera: substitute `VideoCapturer` with a stub that returns synthetic frames, run Qt under `QT_QPA_PLATFORM=offscreen`, open the live preview and then close the main window.

| | before | after |
|---|---|---|
| exit | `QThread: Destroyed while thread '' is still running`, killed by signal | exit code 0 |
| camera | never released | released before shutdown |

The new random-face URL was verified live: `content-type: image/jpeg`, 1024×1024.

## Environment

macOS 27.0, Apple M5 Pro, Python 3.14.1, PySide6 6.11.2, onnxruntime 1.28.0, CoreML execution provider.

The fatal path is not platform-specific — any provider can exceed the old 2s worker timeout — but CoreML hits it on every quit because model compilation is slower on first use.

## Checklist status

Verified: shutdown path (headless reproduction above), random-face fetch, both `imread_unicode` call sites, module import and syntax.

Not yet exercised: the full CONTRIBUTING checklist (enhancer on/off, map faces, camera listing, 15-minute soak). These changes only touch teardown and error paths, not the per-frame pipeline, so FPS and boot time are unaffected — but I have not run the soak test.

## Summary by Sourcery

Harden live-preview shutdown and source-image handling to prevent Qt aborts and processing failures.

Bug Fixes:
- Prevent application crashes during live-preview shutdown by ensuring all preview worker threads are stopped and joined before Qt teardown.
- Handle partially initialized preview windows safely when camera startup fails.
- Fix random-face downloads to use the image endpoint, validate image responses, preserve the existing source on failed downloads, and avoid assigning invalid paths.
- Report unreadable source images instead of allowing preview processing threads to crash, while allowing failed sources to be retried.

Enhancements:
- Improve live-preview worker shutdown handling by waiting through slow model initialization and releasing the camera only after workers have exited.

Tests:
- Add offscreen Qt regression coverage for worker shutdown, failed camera initialization, clean process exit, source reloading, and random-face download integrity.


ozp3:
Good catch — fixed in e06a65f.

`last_source_path` was assigned before `imread_unicode()` had succeeded, so a source that failed to read once was never retried: the worker kept `source_image = None` and the guard `source_path != last_source_path` stayed false until the user picked a *different* file. Re-selecting the same path did not recover it either.

`last_source_path` now advances only after a successful read, so the next frame retries the same path. To keep that from flooding the status line at frame rate, the message is reported once per path rather than once per frame, and reset once the read succeeds.

Verified headlessly (stub `VideoCapturer`, `QT_QPA_PLATFORM=offscreen`, models warmed the way `MainWindow` warms them before opening the preview) with a source file that starts as unreadable bytes and is replaced with a valid image mid-run:

| | result |
|---|---|
| retry while unreadable | 42 read attempts over 1.5s (previously 1) |
| status messages | 1, not one per frame |
| recovery after the file became valid | picked up on the next frame |

The still-preview path (`refresh_frame`) keeps its early `return`; it is a one-shot UI call with no cached state to go stale.

kier007:
not ready yet, 10 tests passed and three regression tests failed in `modules/ui.py`, workers retained in `_STRAY_WORKERS` still crash on exit after the 30-second timeout, selecting valid source A → unreadable B → A leaves swapping disabled because the cached path prevents reloading A, and a failed second random-face download overwrites the active image before validation, please finish worker shutdown before exiting, invalidate the cached path when clearing its source, and validate downloads in a separate file before adopting them, the 15-minute OBSBOT/CUDA soak completed without capture failures or a normal-shutdown crash, all four configurations passed fixed-image GPU checks with throughput within about 0.5% of the base, combined mapping and enhancement still needs live-face verification because no usable face was detected during that stage

ozp3:
Thanks — all three were real. Fixed in 3d741ac, with a regression test for each in `tests/test_ui_live_preview.py`.

**Worker shutdown now finishes.** You were right that parking a straggler in `_STRAY_WORKERS` does not avoid the abort, it only moves it to interpreter exit, where the list is torn down too. That container is gone. `closeEvent` waits out a grace period (`WORKER_SHUTDOWN_GRACE_MS`, there to absorb a slow first-time model load quietly), prints that it is still waiting, and then keeps waiting. It terminates because both loops re-check the stop flag between bounded operations — a camera read, a 50 ms queue get, one inference. The capture is released only after both threads are joined; releasing it while the capture thread may still be inside `cap.read()` is not safe, so the ordering matters.

**The cached path is cleared with its image.** A → unreadable B → A is fixed. Clearing `source_image` while keeping `last_source_path` meant the guard `source_path != last_source_path` stayed false when you went back to A, so A was never reloaded and swapping stayed off until a third file was picked.

**Downloads are staged.** The fetch now writes to `<temp>.part`, validates there — actually decoded, not just content-type checked — and only then `os.replace`s it into position. The staging file is removed if anything fails. Writing straight to the shared temp path was what let a bad second fetch destroy the image already in use.

**Tests.** Six tests, against the real widgets and worker threads under `QT_QPA_PLATFORM=offscreen`, with detection and the frame-processor list stubbed so no model is loaded. 2.5s for the file; 11 tests pass together with the two existing suites.

Checked against the parent commit, the three that correspond to your findings fail there and pass here:

```
FAIL: test_switching_back_to_a_working_source_reloads_it
      A was not reloaded after an unreadable source; the cached path kept swapping disabled
FAIL: test_failed_download_leaves_previous_image_intact
      b'<!DOCTYPE html><html>not an image</html>' != b'\xff\xd8\xff\xe0\x00\x10JFIF...'
ERROR: test_close_event_leaves_no_running_worker
```

One caveat on that third one, so it is not oversold: it pins the new contract and errors on the parent commit only because `WORKER_SHUTDOWN_GRACE_MS` does not exist there. Reproducing the old failure honestly needs a worker stuck past the hardcoded 30s, which is not something I wanted to put in a test. `ShutdownExitCodeTest` covers the property you actually care about end to end — it runs the teardown in a subprocess with a worker that outlives the grace period and asserts the process exits 0 with no `QThread: Destroyed` in its output.

Thanks also for the soak and the GPU throughput numbers. Combined mapping + enhancement live-face verification is still open; I have not been able to exercise that path here.

